PR #16, opened 2021-08-17
Enable validation hoook to test config before upgrading/installing

namm2:
I'd like to add a pre-install,pre-upgrade hooks to run vector validate on provided configs.

Fixes: https://github.com/timberio/helm-charts/issues/14

### DIFF
--- a/charts/vector-agent/templates/validate.yaml
+++ b/charts/vector-agent/templates/validate.yaml
@@ -1,0 +1,68 @@
+{{- if .Values.validateConfig }}
+{{- if (empty .Values.existingConfigMap) -}}
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "libvector.configMapName" . }}-test
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-weight": "5"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+  labels:
+    {{- include "libvector.labels" . | nindent 4 }}
+    tests: "true"
+data:
+  {{- if .Values.customConfig }}
+  vector.yaml: |
+{{ tpl (toYaml .Values.customConfig) . | indent 4 }}
+  {{- end }}
+
+{{- end }}
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ include "libvector.fullname" . }}-test
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-weight": "5"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+  labels:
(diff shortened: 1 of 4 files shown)


namm2:
This validation hook will have trouble (failed) if there's environment variable is set insides the config, e.g.:
```yaml
customConfig:
  sinks:
    humio:
      type: humio_logs
      token: ${HUMIO_TOKEN}
```

If I provide `env` vars to the pod template, it'll still be failed because the ENV var is unknown if it's pre-install hook, but it'll be ok in pre-upgrade hook. So I removed the `env` in pod template at the moment.

spencergilbert:
> If I provide `env` vars to the pod template, it'll still be failed because the ENV var is unknown if it's pre-install hook, but it'll be ok in pre-upgrade hook. So I removed the `env` in pod template at the moment.

I think this is going to be problematic. Some `validate` calls will require the environment variables to succeed, for example a component with a healthcheck that requires authentication.

We could pass `--no-environment` which would ignore healthchecks (which could be valuable to exercise) and just validate that the configuration is valid. What do you think?

Also, I'm sorry I failed to link the DCO app to the new repo - would you mind signing your commits? The main repo has a script that can sign multiple: https://github.com/timberio/vector/blob/master/scripts/signoff.sh

namm2:
> > If I provide `env` vars to the pod template, it'll still be failed because the ENV var is unknown if it's pre-install hook, but it'll be ok in pre-upgrade hook. So I removed the `env` in pod template at the moment.
> 
> I think this is going to be problematic. Some `validate` calls will require the environment variables to succeed, for example a component with a healthcheck that requires authentication.
> 
> We could pass `--no-environment` which would ignore healthchecks (which could be valuable to exercise) and just validate that the configuration is valid. What do you think?
> 
> Also, I'm sorry I failed to link the DCO app to the new repo - would you mind signing your commits? The main repo has a script that can sign multiple: https://github.com/timberio/vector/blob/master/scripts/signoff.sh
> 
> EDIT: it seems like a good thing to enable on (at least) the config focused `ci/*` files


Passing `--no-environment` will not validate remap configs which is the main purpose here.

I'll run the signoff script in later commit.

spencergilbert:
> Passing `--no-environment` will not validate remap configs which is the main purpose here.

Ah  - yes, good point... that seems a little strange to me, I'll investigate to see if that's intended (and why).

spencergilbert:
So I can see a couple of possible options:

1. Have the env vars present at the time of the validate job (more pre-install hooks?)
2. Have the validate step be run in an initContainer (optionally)
3. Include the validate call as part of the entrypoint

namm2:
Option 1 works well for my case where the ENV var is an entry inside a secret like this 
 https://github.com/timberio/vector/pull/8638 then it only needs to create a secret in pre-install hook. But this option won't work in case the ENV var is referencing to any else secret/cm.

PR #24, opened 2021-09-01
Experimental Vector chart

### LINKED ISSUE #7: Support maxUnavailable as a percentage in the vector-agent Helm Chart
@sidewinder12s commented on [Mon Jul 26 2021](https://github.com/timberio/vector/issues/8461)

### Current Vector Version

```
master
```

### Use-cases

Support maxUnavailable as a percentage in the vector-agent Helm Chart. Right now it only supports integers.

### Attempted Solutions

Tried to set a percentage, does not work.

### Proposal

Update the maxUnavailable parameter to support percentages (most likely just need to ensure we use a string) or remove it and just make the entire rolling update config values that users pass in.

### References

Could do something similar to what the aws-vpc-cni plugin has done: https://github.com/aws/eks-charts/blob/master/stable/aws-vpc-cni/templates/daemonset.yaml#L9

### LINKED ISSUE #8: Update vector-agent helm chart to allow Deployment instead of DaemonSet
@aliusmiles commented on [Thu Jun 24 2021](https://github.com/timberio/vector/issues/8019)

### Current Vector Version

```
0.14.0
```

### Use-cases

Vector agent reading AWS ALB logs from S3 bucket and pushing them to Loki.

### Attempted Solutions

vector-agent running as scalable Deployment instead of DaemonSet.

### Proposal

Right now Helm chart for vector-agent uses DaemonSet and is rather hardcoded for only one role: collecting container logs from k8s nodes. In my case Promtail is already doing that so I want to use Vector to only get LB logs from s3 bucket into Loki. I've already tested that locally with minimal effort and minimal config, which is just awesome (especially out-of-the-box parse_aws_alb_log transformer) :-) But when it got to deployment I've realized official chart won't suite my case at all: it's possible to add custom config and disable default sources and sinks, but running vector as DaemonSet for that just doesn't make sense. 

I imagine this might be quite common case - using vector for collecting logs from sources other than container logs or even running multiple agents - one for container logs as daemon, another for s3 logs as single replica, others for something else as auto-scalable deployment.

I remember seeing some chart with had templates for both DaemonSet and Deployment, controlled by variable like `deploymentType: daemonset|deployment`. Would be nice to have something like this in vector-agent chart. Should be relatively simple to introduce additional variable and template, although some extra changes might be needed, for example: Deployment for collecting logs from s3 does not really need /var/log mounted from k8s node, but still might need volume for data-dir, if type is set to Deployment then optionally add hpa, etc.

I believe such feature would add much more flexibility to official chart and allow it to be used in more scenarios. If anybody is interested I can try to draft a PR for that. 

---

@spencergilbert commented on [Thu Jun 24 2021](https://github.com/timberio/vector/issues/8019#issuecomment-867754991)

Hey! I think your use case would be suited for our [vector-aggregator chart](https://github.com/timberio/vector/tree/master/distribution/helm/vector-aggregator). We're currently working on getting it ready to leave beta, but it's definitely usable in it's current state.

Had you seen that chart yet?

---

@aliusmiles commented on [Thu Jun 24 2021](https://github.com/timberio/vector/issues/8019#issuecomment-867955019)

Just checked it out, I should be able to use it for my case, thanks! Just need to disable default sources/sinks/service and add custom config. I have actually saw that chart earlier but haven't looked inside as I was under impression that it can only aggregate data from vector agents.

The only thing that it's a StatefulSet, so no hpa. Doesn't matter for me but if someone needs to get a lot of logs from s3 or other source, I believe deployment with hpa would've been better. 

Possible solution: just add yet another chart for 'vector-forwarder' or something? :) basically vector-agent without host volumes but with deployment and hpa? Though I'm not sure how much demanded that would be, maybe leave this issue open for a while to see if there will be any upvotes? 

My personal case should be solved with aggregator chart I think, will try tomorrow. Thanks for the suggestion!

---

@spencergilbert commented on [Thu Jun 24 2021](https://github.com/timberio/vector/issues/8019#issuecomment-867959422)

👍 @aliusmiles that's definitely a good point. At risk of supporting too many charts that are _similar_ but a little different it would be nice to support HPAs. We did originally choose StatefulSet for the persistent volumes.

I'm happy to leave this open to see if there's general interest in a more "ephemeral, but scalable" strategy. Or maybe HPAs will support StatefulSets one day 😄 

---

@aliusmiles commented on [Fri Jun 25 2021](https://github.com/timberio/vector/issues/8019#issuecomment-868640346)

aggregator chart worked quite nicely for me, thanks! 

One minor suggestion: maybe wrap headless-service template in the same condition as service? Since I've disabled default source, I don't have service in my deployment but chart still deployed headless service with no ports. Not a big issue but still.

Also, is there any issue or something to keep track on when vector-aggregator chart will be by fully released?

P.S. Here's my configs in case someone get interested:

<details>
  <summary>Configs</summary>
  
  values.yaml
  ```yaml
  replicas: 1

  image:
    tag: "0.14.0-alpine"

  nameOverride: "vector"

  serviceAccount:
    create: true
    annotations:
      eks.amazonaws.com/role-arn: IRSA_ROLE_ARN

  tolerations: []

  storage:
    mode: managedPersistentVolumeClaim
    managedPersistentVolumeClaim:
      size: 10Gi

  vectorApi:
    enabled: false
  vectorSource:
    enabled: false
  internalMetricsSource:
    enabled: false
  prometheusSink:
    enabled: false

  haproxy:
    enabled: false

  sources:
    s3_alb_logs:
      type: "aws_s3"
      region: "AWS_REGION"
      compression: "gzip"
      sqs:
        queue_url: "SQS_QUEUE_URL"

  transforms:
    alb_logs:
      type: "remap"
      inputs:
        - "s3_alb_logs"
      source: " . = parse_aws_alb_log!(.message)"

  sinks:
    loki:
      type: "loki"
      inputs:
        - "alb_logs"
      endpoint: "http://loki:3100"
      encoding:
        codec: "json"
      labels:
        forwarder: "vector"
        event: "{{ event_field }}"
        source: "alb_logs"
  ```
  
  helmfile.yaml
  ```yaml
  repositories:
    - name: timberio
      url: git+https://github.com/timberio/vector@distribution/helm?ref=master

  releases:
    - name: vector
      chart: timberio/vector-aggregator
      namespace: monitoring
      values:
        - envs/{{ .Environment.Name }}/values/vector.yaml
  ```
</details>

---

@spencergilbert commented on [Fri Jun 25 2021](https://github.com/timberio/vector/issues/8019#issuecomment-868645673)

@aliusmiles thanks! I'll take a look at that and open an issue based on what I find.

I don't have a hard date for that yet, but it's one of our largest focuses. We'll definitely do a blog post/highlight/tweet/etc when it happens :)

### LINKED ISSUE #18: Use helm-docs to generate README files
https://github.com/norwoodj/helm-docs

### LINKED ISSUE #21: Update aggregator default configuration
We should upgrade the aggregator default configuration to cover a wider range of common sources. This will improve onboarding and simplify installation/quick start.

We should ensure that overriding these defaults is easy for operators that want granular control of their config.

spencergilbert:

Experiment with a single chart that may be configured as either an agent or aggregator, more closely aligning with how we view Vector internally.

This also minimizes dependencies (no vector-shared) while still sharing logic between both roles.

- [x] Closes #18
  - Add similar [docs check](https://github.com/DataDog/helm-charts/blob/main/.github/helm-docs.sh)
- [x] Closes #21 
- [x] Closes #7 
- [x] Closes #8

TODO:

- [ ] Finish README.md
  - [ ] Document upgrade paths
- [ ] Ensure parity with agent
  - [ ] Ensure adequate test coverage with `**/ci/*-values.yaml` files
- [ ] Ensure parity with aggregator
  - [ ] Ensure adequate test coverage with `**/ci/*-values.yaml` files

I suspect that this would be more easily reviewed as multiple PRs (moving through a couple of pre-release versions)


charts/vector-experimental/templates/_pod.tpl
```diff
@@ -0,0 +1,140 @@
...
+{{- with .Values.affinity }}
+affinity:
+{{ toYaml . | indent 2 }}
+{{- end }}
+{{- with .Values.tolerations }}
+tolerations:
+{{ toYaml . | indent 2 }}
+{{- end }}
+volumes:
+{{- if and .Values.persistence.enabled }}
```
hkaj, line 118: there's something strange here, what do we want to do if persistence is enabled?

charts/vector-experimental/templates/daemonset.yaml
```diff
@@ -0,0 +1,23 @@
+{{- if (eq .Values.role "Agent") -}}
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: {{ include "vector.fullname" . }}
```
hkaj, line 5: Do you think we could make the namespace configurable? Pretty sure users will request it. If we do we should also include the namespace in the other k8s resources the chart creates.

charts/vector-experimental/templates/_pod.tpl
spencergilbert: Ah - handled fully fleshed out the persistence options. Definitely needs more work, but was just getting some toggle between PVCs and emptyDirs so `ct` would pass.

charts/vector-experimental/templates/daemonset.yaml
spencergilbert: That's always something that interested me - I always felt like it didn't need to specified in templates because you can just pass `--namespace` to the helm install/upgrade commands. Maybe I'm missing something?

charts/vector-experimental/templates/daemonset.yaml
```diff
@@ -0,0 +1,23 @@
...
+      {{- include "vector.selectorLabels" . | nindent 6 }}
+  template:
+    metadata:
+      {{- with .Values.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      labels:
+        {{- include "vector.selectorLabels" . | nindent 8 }}
+        vector.dev/exclude: "true"
```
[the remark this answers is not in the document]
spencergilbert: Ported directly from the existing chart - I believe generally used to avoid collecting the agent logs if it's _also_ writing what it's collected back to stdout with a `console` sink (infinite loops and all). Definitely could be optional though.

charts/vector-experimental/templates/daemonset.yaml
hkaj: Yeah actually i remembered we had it in the agent chart, but we just set it to `{{ .Release.Namespace }}` so it's not really useful, and we rely on `--namespace` as well 👌

spencergilbert: Cool - it always seemed like an unneeded line with the CLI option (other than say RoleBindings that need the namespace outside of the metadata)

charts/vector-experimental/templates/daemonset.yaml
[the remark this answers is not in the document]
spencergilbert: It actually has come up once - https://github.com/timberio/helm-charts/issues/6 so there's at least _some_ desire :)

charts/vector-experimental/templates/rbac.yaml
```diff
@@ -0,0 +1,31 @@
...
+metadata:
+  name: {{ include "vector.fullname" . }}
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - namespaces
+      - pods
+    verbs:
+      - watch
```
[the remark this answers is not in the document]
hkaj: Gotcha thank you. and in agent mode, vector is expected to run on all nodes or just a small subset? I'm not saying we need to change this right now of course, I suppose that would be a large undertaking, but we may wanna chat about the design of this source to see if we can have a low-impact implementation similar to the kubernetes logs collection in the datadog agent. Especially if we create as many watchers as there are nodes in the cluster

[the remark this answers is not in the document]
spencergilbert: The current expectation is an on-each-node with direct access to the log files. I definitely agree we could use a chat about it, we have had some thoughts about remodeling it entirely - but hasn't been on a nearterm roadmap due to the focus on the aggregator role

binarylogic (approved):
Just noting. I'm a fan as long as users can still specify the `agent` or `aggregator` role and have the defaults adjust accordingly.

spencergilbert:
> Just noting. I'm a fan as long as users can still specify the `agent` or `aggregator` role and have the defaults adjust accordingly.

👍 the current specification via options (or values file) would be `--set role=Agent|Aggregator` (I think capitalized only for now, but could handle normalization)

charts/vector-experimental/templates/daemonset.yaml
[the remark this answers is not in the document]
spencergilbert: Going to punt on this until I have a better solution - perhaps updated defaults. Writing logs to stdout + an agent collecting is 😭

charts/vector/templates/NOTES.txt
```diff
@@ -0,0 +1,26 @@
...
+  \_/  \/
+V E C T O R
+
+Vector is starting in your cluster. After a few minutes, you can use Vector's API to view internal metrics by running:
+{{ if (eq .Values.role "Aggregator") }}
+  {{- if not .Values.customConfig }}
+	kubectl -n {{ .Release.Namespace }} exec -it statefulset/{{ include "vector.fullname" . }} -- vector top 
+  {{- else }}
+	kubectl -n {{ .Release.Namespace }} exec -it statefulset/{{ include "vector.fullname" . }} -- vector top --url {{ .Values.customConfig.api.address }}
+  {{- end }}
```
jszwedko, line 13: Nit, it seems like these are all the same besides the prefix. I might suggest pulling out the calculation of the pod name somewhere where it can be shared (if possible).

jszwedko (approved):
Looks good! I mostly just scanned the actual config for each role since I assumed it was just copy/pasted from the current agent and aggregator roles. Let me know if there is anything in particular you want me to 👀 at though.

charts/vector/ci/default-agent-values-BROKEN.yaml
```diff
@@ -0,0 +1,7 @@
+# Agent role
+## Values file for testing default Agent parameters.
+role: Agent
+
+# This is currently broken without persistence enabled.
+# It appears to be related to us mounting `/var/lib` from
+# the host but the host not having an exiting `/var/lib/vector`.
```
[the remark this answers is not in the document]
spencergilbert: Yep - I'm not _entirely_ sure why we set that VOLUME in the dockerfile (brought up in slack last night 🦗 ). I believe 1:1 from the old charts would just be _always_ setting a hostPath for agents, and not allowing "non-persistent" agents.

charts/vector/templates/NOTES.txt
spencergilbert: Indeed - there are a number of places I expect to DRY before (or after) an actual release, checking for role is another. Since it's not necessarily user facing I deferred while I worked on actual functionality (at least until the point it got really annoying to work with 😉 )

charts/vector/ci/default-agent-values-BROKEN.yaml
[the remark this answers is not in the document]
spencergilbert: Discussed in slack - next release will just enforce a hostPath for this path when deploying as an Agent

PR #28, opened 2021-09-03
Create a single check that can be used for PRs

spencergilbert:

```yaml
  pr-validated:
    name: pr-validated
    runs-on: ubuntu-latest
    needs:
      - install-chart
    steps:
      - name: validate
        run: echo "PR OK"
```

Added this from Datadog/helm-charts that can be used as a single check for PRs (otherwise each matrix job needs to be selected)


jszwedko (approved):
Ah yes, I think we do this somewhere in vector too for matrix jobs.

PR #41, opened 2021-09-10
spencer/vector alpha.2

spencergilbert:
- fix: Add -0 to kubeVersion to fix comparison on EKS
- fix: Render env in pod template
- chore: Document podAnnotation option
- chore: Improve NOTES output accuracy
- feat: Document and implement serviceaccount options
- feat: Add checksum of generated configmap for all roles
- feat: Add podLabels for all workloads
- feat: Add priorityClassName for all workloads
- chore: Minor comments rewrites
- feat: Add pullSecrets to pod template
- feat: Set Vector args through values file
- feat: Allow for manual overriding of Service ports
- chore: Update docs

### DIFF
--- a/charts/vector/Chart.yaml
+++ b/charts/vector/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: vector
-version: "0.1.0-alpha.1"
-kubeVersion: ">=1.15.0"
+version: "0.1.0-alpha.2"
+kubeVersion: ">=1.15.0-0"
 description: A lightweight, ultra-fast tool for building observability pipelines
 type: application
 keywords:
--- a/charts/vector/templates/NOTES.txt
+++ b/charts/vector/templates/NOTES.txt
@@ -4,23 +4,34 @@ __   __  __
   \_/  \/
 V E C T O R
 
-Vector is starting in your cluster. After a few minutes, you can use Vector's API to view internal metrics by running:
 {{ if (eq .Values.role "Aggregator") }}
   {{- if not .Values.customConfig }}
+Vector is starting in your cluster. After a few minutes, you can use Vector's API to view internal metrics by running:
+
 	kubectl -n {{ .Release.Namespace }} exec -it statefulset/{{ include "vector.fullname" . }} -- vector top 
-  {{- else }}
-	kubectl -n {{ .Release.Namespace }} exec -it statefulset/{{ include "vector.fullname" . }} -- vector top --url {{ .Values.customConfig.api.address }}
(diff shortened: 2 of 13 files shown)


charts/vector/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: vector
-version: "0.1.0-alpha.1"
-kubeVersion: ">=1.15.0"
+version: "0.1.0-alpha.2"
+kubeVersion: ">=1.15.0-0"
```
spencergilbert, line 4: This would fail when attempting to install against an EKS cluster:

```shell
Error: UPGRADE FAILED: chart requires kubeVersion: >=1.15.0 which is incompatible with Kubernetes v1.20.7-eks-d88609
```

charts/vector/templates/NOTES.txt
```diff
@@ -4,23 +4,34 @@ __   __  __
   \_/  \/
 V E C T O R
```
spencergilbert, line 6: Still need to refactor this out into some `include` or `template` but I caught some UX issues in the meantime

charts/vector/templates/_pod.tpl
```diff
@@ -15,10 +22,14 @@ containers:
 {{- end }}
     image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
     imagePullPolicy: {{ .Values.image.pullPolicy }}
+{{- with .Values.args }}
     args:
-      - --config-dir
-      - "/etc/vector/"
+    {{- toYaml . | nindent 6 }}
```
spencergilbert, line 27: Allows for users to add more options/flags or replace our default config dirs (pipelines soon). This should be less brittle as a user needs to replace each item in the list, so us changing defaults wouldn't impact someone already adjusting the args (vs appending additional args to the end of the list)

charts/vector/templates/service-headless.yaml
```diff
@@ -5,10 +5,16 @@ metadata:
...
+  {{- with .Values.service.annotations }}
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
 spec:
   clusterIP: None
   ports:
-{{- if .Values.customConfig }}
+{{- if .Values.service.ports }}
+  {{- toYaml .Values.service.ports | nindent 4 }}
+{{- else if .Values.customConfig }}
```
spencergilbert, line 17: the headless service is mirroring the normal service, which may be ok - but might need to introduce more granular control...

StephenWakely (approved):
Looks good to me!

jszwedko (approved):
🚀

PR #45, opened 2021-09-10
chore: Add features from vector-aggregator chart

### LINKED ISSUE #35: Ensure feature parity between vector and vector-aggregator
Ensure adequate test coverage with `**/ci/*-values.yaml` files

From #24

spencergilbert:
- feat(vector): Add dnsConfig and dnsPolicy support
- feat(vector): Add support for an existing ConfigMap
- feat(vector): Add support for termination period and management policy
- feat(vector): Add PodMonitor and update default prometheus port name
- feat(vector): Support existing ConfigMap for HAProxy

Closes #35

I still need to determine what options we should exercise in CI for the Aggregator.


spencergilbert:
I _think_ I'll update the `ci` files to be `<role>-<"purpose">-values.yaml` like `aggregator-customConfig-values.yaml`

charts/vector/templates/haproxy/configmap.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.haproxy.enabled }}
+{{- if and .Values.haproxy.enabled (not .Values.existingConfigMap) }}
```
StephenWakely, line 1: Does `not` on an empty string return `true`?

spencergilbert: 👍  empty is ~= false when evaluating templates

charts/vector/templates/_pod.tpl
```diff
@@ -73,11 +80,11 @@ containers:
       - name: vector
         containerPort: 6000
         protocol: TCP
-      - name: prometheus
+      - name: prom-exporter
```
StephenWakely, line 83: Here `prometheus` is being renamed to `prom-exporter`, but below `prometheus-exporter` is being renamed to `prom-exporter`. Is it ok that things that had different names now have the same name?

spencergilbert: So I found out that `containerPort` names can't be more than 15(16?) characters, which broke when autogenerating them based off `customConfig` - I wanted to consistently name things across `Pods` and `Services` (as well as the provided defaults and what the generated output is.

tl;dr I wanted to ensure that if you passed what I've provided as default into the `customConfig` option you'd get the same output as just using the defaults.

StephenWakely (approved):
Looks good to me!

charts/vector/templates/NOTES.txt
```diff
@@ -12,7 +12,7 @@ Vector is starting in your cluster. After a few minutes, you can use Vector's AP
   {{- else if .Values.customConfig.api.enabled }}
 Vector is starting in your cluster. After a few minutes, you can use Vector's API to view internal metrics by running:
 
-	kubectl -n {{ .Release.Namespace }} exec -it statefulset/{{ include "vector.fullname" . }} -- vector top --url {{ .Values.customConfig.api.address | default "http://127.0.0.1:8686/graphql" }}
+	kubectl -n {{ .Release.Namespace }} exec -it statefulset/{{ include "vector.fullname" . }} -- vector top --url {{ printf "http://%s/graphql" .Values.customConfig.api.address | default "http://127.0.0.1:8686/graphql" }}
```
spencergilbert, line 15: This command would fail if the `api.address` was replaced in `customConfig` as we didn't pass the full url and path

jszwedko (approved):
Looks good! Thanks for keeping the commits specific.

PR #58, opened 2021-09-21
feat(vector)!: Support injecting multiple ConfigMaps

spencergilbert:
- feat(vector): User projected volume for config flexibiltiy
- feat(vector): Allow extra ConfigMaps to be added to Vector's config dir
- chore(vector): Rename existingConfigMap options for consistency


charts/vector/templates/_pod.tpl
```diff
@@ -149,8 +149,16 @@ volumes:
...
+    projected:
+      sources:
+        - configMap:
+            name: {{ if .Values.existingConfig }}{{ .Values.existingConfig }}{{ else }}{{ template "vector.fullname" . }}{{ end }}
+{{- with .Values.extraConfigs }}
+  {{- range . }}
+        - configMap:
+            name: {{ . }}
+  {{- end }}
+{{- end }}
```
spencergilbert, line 161: The benefit of using `projected` volumes is that you can group a number of sources (ConfigMaps, Secrets, etc) into a single volume, which makes things neater in the templating.

With a `projected` volume rather than ranging and creating `n` volumes for each configmap, we have one volume with `n` configmaps (or secrets if needed).

It's a cleaner abstraction for what we're trying accomplish.

charts/vector/README.md
```diff
@@ -92,7 +92,8 @@ helm install --name <RELEASE_NAME> \
 | dnsConfig | object | `{}` | Specify DNS configuration options for Vector Pods |
 | dnsPolicy | string | `"ClusterFirst"` | Specify DNS policy for Vector Pods |
 | env | list | `[]` | Set environment variables in Vector containers |
-| existingConfigMap | string | `""` | Use existing ConfigMap for Vector's configuration instead of creating a new one |
+| existingConfig | string | `""` | Use this existing ConfigMap for Vector's configuration instead of creating a new one |
```
spencergilbert, line 95: The rename here is for two reasons:

1. Better matching the `extraConfigs`
2. Leaving space for an eventual `pipelines` option.

Both of which are just naming, I couldn't think of something that fit with the existing nicely like `extraConfigConfigMaps` (because one day an `extraPipelinesConfigMaps` will probably be needed). Then the same for eventual `existingPipelineConfigMap` is _fine_ish but leaves `existingConfigMap` more vague and not matching the "naming scheme"

jszwedko (approved):
Looks like there is a CI failure, but the changes look good to me!

PR #65, opened 2021-09-30
chore: Document nameOverride and fullnameOverride

spencergilbert:

These were used in templates but not documented nor included in the values file.

### DIFF
--- a/charts/vector/README.md
+++ b/charts/vector/README.md
@@ -95,6 +95,7 @@ helm install --name <RELEASE_NAME> \
 | env | list | `[]` | Set environment variables in Vector containers |
 | existingConfig | string | `""` | Use this existing ConfigMap for Vector's configuration instead of creating a new one, if used requires dataDir to be set |
 | extraConfigs | list | `[]` | List of ConfigMap names to include as additional configuration files |
+| fullnameOverride | string | `""` | Override the full qualified app name |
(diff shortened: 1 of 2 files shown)


charts/vector/README.md
```diff
@@ -105,6 +106,7 @@ helm install --name <RELEASE_NAME> \
 | ingress.hosts | list | `[]` | Configure the hosts and paths for the Ingress |
 | ingress.tls | list | `[]` | Configure TLS for the Ingress |
 | livenessProbe | object | `{}` | Override default liveness probe settings, if customConfig is used requires customConfig.api.enabled true |
+| nameOverride | string | `""` | Override name of app |
```
StephenWakely, line 109:
```suggestion
| nameOverride | string | `""` | Override the name of the app |
```

StephenWakely (approved):
Nice.

My only question, do kubernetes people automatically know what a fully qualified name is, or is that something that might be worth explaining?

spencergilbert:
> Nice.
> 
> My only question, do kubernetes people automatically know what a fully qualified name is, or is that something that might be worth explaining?

@StephenWakely I'd say it's common, these are created with any `helm create`d chart and would be seen with almost every chart usage.

I copied both of the values and descriptions from the datadog chart actually 😄

PR #73, opened 2021-09-30
feat(vector): Add parameter to include additional labels on all resources

spencergilbert:

This paramter wasn't ported from the old charts, I noticed it while writing the upgrade guides


StephenWakely (approved):
Seems legit.

PR #74, opened 2021-10-01
feat(vector): Add extraVolumes and extraVolumeMounts parameters

spencergilbert:

Added from the old charts


StephenWakely (approved):
Looks good!

Is there a way to make it so that it checks that the `extraVolume`s correspond to the `extraVolumeMounts` specified? I guess the container runtime will check that before installing anyway, so there's no need for us to do so..

spencergilbert:
> Looks good!
> 
> Is there a way to make it so that it checks that the `extraVolume`s correspond to the `extraVolumeMounts` specified? I guess the container runtime will check that before installing anyway, so there's no need for us to do so..

🤔  we didn't in the past, and I _think_ there may be times that it's possible to define mounts without the corresponding volume (if something else is defining it), I think it'd be ok to address it in the future if it comes up.

Issue #33, opened 2021-09-08
Finish vector README


spencergilbert:
Document upgrade path for vector-agent and vector-aggregator

From #24

PR #75, opened 2021-10-01
docs(vector): Add upgrade docs

spencergilbert:
Part of #33

### DIFF
--- a/charts/vector/README.md
+++ b/charts/vector/README.md
@@ -1,6 +1,6 @@
 # Vector
 
-![Version: 0.1.0-alpha.4](https://img.shields.io/badge/Version-0.1.0--alpha.4-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 0.16.1-distroless-libc](https://img.shields.io/badge/AppVersion-0.16.1--distroless--libc-informational?style=flat-square)
+![Version: 0.1.0-beta.0](https://img.shields.io/badge/Version-0.1.0--beta.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 0.16.1-distroless-libc](https://img.shields.io/badge/AppVersion-0.16.1--distroless--libc-informational?style=flat-square)
 
 [Vector](https://vector.dev/) is a high-performance, end-to-end observability data pipeline that puts you in control of your observability data. Collect, transform, and route all your logs, metrics, and traces to any vendors you want today and any other vendors you may want tomorrow. Vector enables dramatic cost reduction, novel data enrichment, and data security where you need it, not where is most convenient for your vendors.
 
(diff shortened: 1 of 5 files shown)


charts/vector/README.md
```diff
@@ -32,15 +32,12 @@ helm install --name <RELEASE_NAME> \
 
 ### Upgrading
 
-#### From vector-agent
+#### From original charts
```
Yasumoto, line 36: Aha, just saw this while looking to see if there was an issue/PR filed for https://github.com/vectordotdev/helm-charts/issues/77

I think it'd be super helpful to add a note similar to https://github.com/vectordotdev/helm-charts/pull/76 (feel free to close that) for a little extra context for folks!

```suggestion

The [Vector chart](charts/vector/README.md) is the recommended way to deploy Vector in either the Aggregator or Agent roles. The original charts are deprecated, please see the following migration guides: 
```

spencergilbert: 👍  I wanted to make sure we had the new `vector` chart in a stable (bug free 😅 ) spot before messaging it as a replacement of the original charts. I suppose once the migration docs are in place and this is merged that'll probably be a good time to do so.

As a side-note your suggested change should _technically_ go into `README.md.gotmpl` otherwise `helm-docs` would overwrite it 😃

charts/vector/README.md
```diff
@@ -63,6 +60,34 @@ helm upgrade -f values.yaml <RELEASE_NAME> vector/vector
 
 See the [All configuration options](#all-configuration-options) section to discover all possibilities offered by the Vector chart.
 
+### Running on control plane nodes
```
jszwedko, line 63: 👍

jszwedko (commented):
Nice! Short and sweet. I appreciate that the upgrades are straight-forward and require minimal changes.

Can you remind me of the commands users would actually run to upgrade? Would it just be a `helm install` of the new chart? Or do they need to do anything special?

jszwedko:
We may also want to add a note about this upgrade to the docs in `vectordotdev/vector` after this is released, too.

spencergilbert:
> Nice! Short and sweet. I appreciate that the upgrades are straight-forward and require minimal changes.
> 
> Can you remind me of the commands users would actually run to upgrade? Would it just be a `helm install` of the new chart? Or do they need to do anything special?

I think it should be `helm upgrade <old release name> vector/vector`

jszwedko:
> > Nice! Short and sweet. I appreciate that the upgrades are straight-forward and require minimal changes.
> > Can you remind me of the commands users would actually run to upgrade? Would it just be a `helm install` of the new chart? Or do they need to do anything special?
> 
> I think it should be `helm upgrade <old release name> vector/vector`

Ah, perfect. Does it make sense to include that in the upgrade guide? Or would that be redundant?

spencergilbert:
> Ah, perfect. Does it make sense to include that in the upgrade guide? Or would that be redundant?

I mean 🤷 I initially felt like it was an "obvious" - but I'll do the upgrades this morning and add the exact commands to the two docs

jszwedko (approved):
Thanks for adding the upgrade commands!

PR #76, opened 2021-10-01
docs(vector): Clarify suggested usage

Yasumoto:
These charts used to be hosted within the main repo, and #24 added a combined chart that looks to be the future. This adds a (likely short-term) pointer in the README to help guide folks to the right path.

Maybe most importantly, please consider this a strawman since I'm not entirely certain this is the chart y'all recommend folks use!

This is related to https://github.com/vectordotdev/vector/pull/9396

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,8 @@ Official Helm charts for Vector. Currently supported:
 - [Vector Agents](charts/vector-agent/README.md) (vector/vector-agent)
 - [Vector Aggregators](charts/vector-aggregator/README.md) (vector/vector-aggregator)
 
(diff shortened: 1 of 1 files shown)


spencergilbert:
Thanks! I was actually holding off on the messaging until I moved the `vector` chart out of pre-release (hopefully today/early next week) - but the way you phrased it makes sense to me right now 😄

Yasumoto:
Totally makes sense! As I started to look into the repo and recent changes I figured that was the case 👍🏼 

Love all the changes you're making, it's going to be great!

PR #82, opened 2021-10-01
chore: Set continue-on-error to false in CI

spencergilbert:

Should resolve the issue I ran into where a PR was merged when CI failed

Ref: https://github.community/t/status-check-for-a-matrix-jobs/127354/7

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -119,10 +119,12 @@ jobs:
         run: ct install --config .github/ct.yaml --excluded-charts vector-shared --upgrade
 
   pr-validated:
+    if: ${{ always() }}
     name: pr-validated
     runs-on: ubuntu-latest
     needs:
       - install-chart
     steps:
       - name: validate
-        run: echo "PR OK"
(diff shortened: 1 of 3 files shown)


jszwedko (commented):
I think you need to modify the check step at the bottom to do something like:

```
      - name: Check build matrix status
        if: ${{ needs.build.result != 'success' }}
        run: exit 1
```

Currently, I think your changes here will cause it to run, but won't cause it to fail.

You could introduce a failure here to verify it.

spencergilbert:
> I think you need to modify the check step at the bottom to do something like:
> 
> ```
>       - name: Check build matrix status
>         if: ${{ needs.build.result != 'success' }}
>         run: exit 1
> ```
> 
> Currently, I think your changes here will cause it to run, but won't cause it to fail.
> 
> You could introduce a failure here to verify it.

I'll do some _actual_ reading on how these GHA things work 🙄

jszwedko:
@spencergilbert did this work out? I do think it is valuable for automerge.

PR #91, opened 2021-11-01
Document how to use Vector's template syntax in customConfig

### LINKED ISSUE #87: helm templating chart bug: {{...}} doesn't work in customConfig
Hi when using  `customConfig:` and defining in sinks something like

`key_prefix: "{{ kubernetes.pod_labels.app\\.kubernetes\\.io/client-id }}/%Y/%m/%d/{{ kubernetes.pod_name }}/"`

it fails with: 

> function "kubernetes" not defined

spencergilbert:

Closes #87

### DIFF
--- a/charts/vector/README.md
+++ b/charts/vector/README.md
@@ -63,6 +63,25 @@ helm upgrade -f values.yaml <RELEASE_NAME> vector/vector
 
 See the [All configuration options](#all-configuration-options) section to discover all possibilities offered by the Vector chart.
 
+### Using template syntax in `customConfig`
+
+As Vector's [template syntax](https://vector.dev/docs/reference/configuration/template-syntax/) shares the same syntax as Go templates, explicit handling is required
+if you are using Vector's template syntax in the `customConfig` option. To avoid Helm from templating configuration intended for Vector you can supply configuration like so:
+
+```yaml
+customConfig: |
+  #...
+  sinks:
+    loki:
+      #...
+      labels:
+        foo: bar
+	host: |
+	  {{ host }}
+	source: |
+	  {{ source_type }}
+```
+
 ## All configuration options
 
 The following table lists the configurable parameters of the Vector chart and their default values. Specify each parameter using the `--set key=value[,key=value]` argument to `helm install`. For example,
--- a/charts/vector/README.md.gotmpl
+++ b/charts/vector/README.md.gotmpl
@@ -61,6 +61,25 @@ helm upgrade -f values.yaml <RELEASE_NAME> vector/vector
 
 See the [All configuration options](#all-configuration-options) section to discover all possibilities offered by the Vector chart.
 
+### Using template syntax in `customConfig`
+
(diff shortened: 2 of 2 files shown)


charts/vector/README.md
```diff
@@ -63,6 +63,25 @@ helm upgrade -f values.yaml <RELEASE_NAME> vector/vector
...
+  #...
+  sinks:
+    loki:
+      #...
+      labels:
+        foo: bar
+	host: |
+	  {{ host }}
+	source: |
+	  {{ source_type }}
```
jszwedko, line 82: I think you are mixing tabs and spaces here

charts/vector/README.md
```diff
@@ -63,6 +63,25 @@ helm upgrade -f values.yaml <RELEASE_NAME> vector/vector
 
 See the [All configuration options](#all-configuration-options) section to discover all possibilities offered by the Vector chart.
 
+### Using template syntax in `customConfig`
+
+As Vector's [template syntax](https://vector.dev/docs/reference/configuration/template-syntax/) shares the same syntax as Go templates, explicit handling is required
```
jszwedko, line 68:
```suggestion
As Vector's [template syntax](https://vector.dev/docs/reference/configuration/template-syntax/) shares the same syntax as Helm templates, explicit handling is required
```

Users may not realize that Helm is using Go's templating language since they don't seem to emphasize than in their docs.

jszwedko (commented):
Nice, this should definitely help. I wonder if we should mention it on https://vector.dev/docs/setup/installation/platforms/kubernetes/#helm too?

charts/vector/README.md
spencergilbert: 🤔 happy to change it but: https://helm.sh/docs/howto/charts_tips_and_tricks/#know-your-template-functions

jszwedko: True, but that doc is aimed at chart developers rather than chart users.

charts/vector/README.md
```diff
@@ -63,6 +63,25 @@ helm upgrade -f values.yaml <RELEASE_NAME> vector/vector
 
 See the [All configuration options](#all-configuration-options) section to discover all possibilities offered by the Vector chart.
 
+### Using template syntax in `customConfig`
+
+As Vector's [template syntax](https://vector.dev/docs/reference/configuration/template-syntax/) shares the same syntax as Helm templates, explicit handling is required
+if you are using Vector's template syntax in the `customConfig` option. To avoid Helm from templating configuration intended for Vector you can supply configuration like so:
```
jszwedko, line 69:
```suggestion
if you are using Vector's template syntax in the `customConfig` option. To avoid Helm templating configuration intended for Vector you can supply configuration like so:
```

jszwedko (approved):
🚀

spencergilbert:
Committed your suggestion and added a CI test, assuming it passes I'll merge - thanks @jszwedko

PR #101, opened 2021-11-08
fix(vector): Toggle both service and headless service with service.enabled

### LINKED ISSUE #100: vector chart: vector-headless service is still created when service.enable=false
headless service is still created when service is disabled in values. 

service-headless' template condition is [set](https://github.com/vectordotdev/helm-charts/blob/6fcb9de045f2c0b0b01bf99082413b7574114ed1/charts/vector/templates/service-headless.yaml#L1) to `{{- if or (eq .Values.role "Aggregator") (eq .Values.role "Stateless-Aggregator") -}}` 

while service's template [has](https://github.com/vectordotdev/helm-charts/blob/6fcb9de045f2c0b0b01bf99082413b7574114ed1/charts/vector/templates/service.yaml#L1) `{{- if and .Values.service.enabled }}`

Possible solution:
```yaml
{{- if and .Values.service.enabled (or (eq .Values.role "Aggregator") (eq .Values.role "Stateless-Aggregator")) -}}
```

Use case for having service completely disabled: 

using Vector for getting AWS ALB logs from S3 bucket to Loki or any other pull configuration.

spencergilbert:

Closes #100

This mirrors the condition used for both the service and service-headless.
Headless services are generally more useful for load balancing or being able
to select specific instances from a StatefulSet, but rather than introducing
more configuration this seems simpler unless a user requests more specific options.

### DIFF
--- a/charts/vector/README.md
+++ b/charts/vector/README.md
@@ -1,6 +1,6 @@
 # Vector
 
-![Version: 0.1.0-beta.2](https://img.shields.io/badge/Version-0.1.0--beta.2-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 0.17.3-distroless-libc](https://img.shields.io/badge/AppVersion-0.17.3--distroless--libc-informational?style=flat-square)
+![Version: 0.1.0-beta.3](https://img.shields.io/badge/Version-0.1.0--beta.3-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 0.17.3-distroless-libc](https://img.shields.io/badge/AppVersion-0.17.3--distroless--libc-informational?style=flat-square)
 
 [Vector](https://vector.dev/) is a high-performance, end-to-end observability data pipeline that puts you in control of your observability data. Collect, transform, and route all your logs, metrics, and traces to any vendors you want today and any other vendors you may want tomorrow. Vector enables dramatic cost reduction, novel data enrichment, and data security where you need it, not where is most convenient for your vendors.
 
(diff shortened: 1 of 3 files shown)


jszwedko (commented):
What's the difference between the headless and head(ed?) service?

spencergilbert:
> What's the difference between the headless and head(ed?) service?

https://kubernetes.io/docs/concepts/services-networking/service/#headless-services

tl;dr: `For headless Services, a cluster IP is not allocated, kube-proxy does not handle these Services, and there is no load balancing or proxying done by the platform for them.`

For a concrete example, HAProxy in our chart uses the headless service to do DNS resolutions for each vector instance so it can create it's backends.

charts/vector/templates/service-headless.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if or (eq .Values.role "Aggregator") (eq .Values.role "Stateless-Aggregator") -}}
+{{- if and .Values.service.enabled }}
```
jszwedko, line 1: It seems like the `and` here is superfluous?

charts/vector/README.md
```diff
@@ -1,6 +1,6 @@
 # Vector
 
-![Version: 0.1.0-beta.2](https://img.shields.io/badge/Version-0.1.0--beta.2-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 0.17.3-distroless-libc](https://img.shields.io/badge/AppVersion-0.17.3--distroless--libc-informational?style=flat-square)
+![Version: 0.1.0-beta.3](https://img.shields.io/badge/Version-0.1.0--beta.3-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 0.17.3-distroless-libc](https://img.shields.io/badge/AppVersion-0.17.3--distroless--libc-informational?style=flat-square)
```
jszwedko, line 3: Still wondering if this'll conflict with https://github.com/vectordotdev/helm-charts/pull/104

jszwedko (approved):
Looks good! Thanks again for explaining offline (but online).

charts/vector/templates/service-headless.yaml
spencergilbert: Huh - I put the `and` in the other template too 🤔

PR #116, opened 2021-11-23
fix(vector): typo `podPriorityClassName` & `StatefulSet`

CasparChou:
fix two typo:

* Use `podPriorityClassName` in values just like before in `_pod.tpl`
* Not `Statefulset`, is `StatefulSet` in hpa.yml (it also cause some error).

```
Events:
  Type     Reason          Age                   From                       Message
  ----     ------          ----                  ----                       -------
  Warning  FailedGetScale  2m5s (x275 over 72m)  horizontal-pod-autoscaler  no matches for kind "Statefulset" in group "apps"
  ```


jszwedko (approved):
Thank you @CasparChou !

PR #117, opened 2021-11-23
fix(vector): Switch metric list order in HPA

### LINKED ISSUE #118: ArgoCD always thinks the Vector HPA is out of sync.
Due to the order of the spec.metrics list for the HPA, Argo will consider the HPA permanently out of sync:

Screen Shot 2021-11-23 at 1 13 42 PM

sam-mcbr:
What this PR does / why we need it:

Reorders the list in spec.metrics for all components' HPAs in order to circumvent issue kubernetes/kubernetes#74099. Without this change, GitOps style CD tools such as Argo will consider the HPAs permanently out of sync due to the following discrepancy:

Screen Shot 2021-11-23 at 1 13 42 PM

Fixes #118

### DIFF
--- a/charts/vector/Chart.yaml
+++ b/charts/vector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: vector
-version: "0.2.0"
+version: "0.2.1"
 kubeVersion: ">=1.15.0-0"
 description: A lightweight, ultra-fast tool for building observability pipelines
 type: application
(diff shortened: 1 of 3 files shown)


jszwedko (approved):
That is an odd bug, thanks for the fix @sam-mcbr 

Normally you can leave the version bump out of the PR as we'll bump it during release, but given I'll be cutting a release soon with this bug fix and #116 I think this is fine.

PR #122, opened 2021-11-26
fix(vector-aggregator): Fix the conditions for enable/disable SA on the HAProxy

ctavian:
This PR fix the 'if' condition in the serviceaccount.yaml file for use only the value in haproxy block in the value file (haproxy.serviceAccount.create).

And add a condition 'if' in the deployment of haproxy for add or remove the annotation 'serviceAccount' if the value is 'haproxy.serviceAccount.create' is true or false.

If I understood correctly, the SA is only used for the PSP and it is deprecated from version v1.21.

### DIFF
--- a/charts/vector-aggregator/templates/haproxy/deployment.yaml
+++ b/charts/vector-aggregator/templates/haproxy/deployment.yaml
@@ -27,7 +27,9 @@ spec:
 {{ toYaml .Values.haproxy.podAnnotations | indent 8 }}
       {{- end }}
     spec:
+    {{- if .Values.haproxy.serviceAccount.create }}
       serviceAccountName: {{ include "libvector.serviceAccountName" . }}-haproxy
+    {{- end }}
       terminationGracePeriodSeconds: {{ .Values.haproxy.terminationGracePeriodSeconds }}
       securityContext:
         {{- toYaml .Values.haproxy.podSecurityContext | nindent 8 }}
--- a/charts/vector-aggregator/templates/haproxy/deployment.yaml
+++ b/charts/vector-aggregator/templates/haproxy/deployment.yaml
@@ -27,9 +27,7 @@ spec:
 {{ toYaml .Values.haproxy.podAnnotations | indent 8 }}
       {{- end }}
     spec:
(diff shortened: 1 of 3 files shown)


charts/vector-aggregator/templates/haproxy/deployment.yaml
```diff
@@ -27,7 +27,9 @@ spec:
 {{ toYaml .Values.haproxy.podAnnotations | indent 8 }}
       {{- end }}
     spec:
+    {{- if .Values.haproxy.serviceAccount.create }}
       serviceAccountName: {{ include "libvector.serviceAccountName" . }}-haproxy
+    {{- end }}
```
spencergilbert, line 32: I think it's common to not wrap this in a conditional, see: https://github.com/haproxytech/helm-charts/blob/main/haproxy/templates/deployment.yaml#L52

Do you have any examples of this pattern in other charts?

spencergilbert (changes requested):
Would you also mind bumping the patch version for the chart in the `Chart.yaml`?

charts/vector-aggregator/templates/haproxy/deployment.yaml
ctavian: Hello, yes indeed this is not a very good solution, the problem is the name of the "ServiceAccountName" "-haproxy" in the deployment file.

When we don't want to activate the "ServiceAccount" in the haproxy, it will use the default "default" and add "-haproxy" at the end to give this name: (default-haproxy) and the replica will fail because it cannot find this serviceaccount.

The "default" value is supplied by the "libvector.serviceAccountName" function present in the "vector-shared" graph which has no condition on the creation value (true / false) of a "ServiceAccount"
Suddenly, it necessarily returns "default".

However, it would be interesting to use the haproxy.serviceAccountName function as for the "vector" graphic to correct this problem. :)

spencergilbert: Doh - oh yeah, much better solution 😄 I missed that completely.

spencergilbert (approved):

PR #129, opened 2021-12-14
docs(vector): Clarify usage of secrets.generic and env

### LINKED ISSUE #128: Update example for secrets
A user was confused and thought that the example of AWS credentials we use as an example of passing secrets was indicating that they needed to provide AWS credentials to run the chart. We should:

* Use an agnostic example that is unlikely to be interpreted as needed configuration
* Update the comment to make it clearer that the provided, commented out, configuration is just an example

spencergilbert:

Closes #128

### DIFF
--- a/charts/vector/Chart.yaml
+++ b/charts/vector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: vector
-version: "0.2.2"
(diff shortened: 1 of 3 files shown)


jszwedko (approved):
Thanks for clarifying these!

PR #150, opened 2022-01-12
chore: Release 0.4.0

jszwedko:


spencergilbert (approved):
I've been lazily merging locally, but should probably start doing it this way 🤔

jszwedko:
Aha, yeah, I don't have a strong opinion. Maybe we should document whichever way is preferred?

PR #154, opened 2022-01-24
docs(vector): Fix loki label ends with new line

cybergrind:
Hi

The current example generates Loki labels ending with a newline. This fixes it


spencergilbert (approved):
I swear I remember fixing this, or something just like it.

Thanks for catch and PR!

PR #157, opened 2022-02-01
feat(vector): Upgrade Vector to 0.19.1

jszwedko:
Signed-off-by: Jesse Szwedko <jesse@szwedko.me>


spencergilbert (commented):
I think the vector- charts should be left as is now that they're marked deprecated

jszwedko:
> I think the vector- charts should be left as is now that they're marked deprecated

True, though since this is a patch release, I think it'd be prudent to include it in the last release of this chart. I bumped the versions of the chart, but we can leave it out of Vector version upgrades in the future.

spencergilbert (approved):
I guess that's fine too

PR #158, opened 2022-02-02
chore: Note chart deprecations in READMEs

jszwedko:


spencergilbert (approved):

spencergilbert:
I'm OK with force merging without updating the version, but your call

PR #172, opened 2022-03-14
Autoscale across AZs by default

silversupreme:
I get a feeling that this won't be enough, but this can at least prompt some discussion on what the right way to go is.

Recent customer deployments have shown that this would be reasonable to use by default, since it's what folks doing production deployments are going to want anyways and I've had to copy/paste this snippet a few times.

### DIFF
--- a/charts/vector/values.yaml
+++ b/charts/vector/values.yaml
@@ -178,7 +178,19 @@ tolerations: []
 
 # affinity -- Allow Vector to schedule using affinity rules
 ## Ref: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity
-affinity: {}
+affinity:
+  ## Scale across different AZs by default.
+  podAntiAffinity:
+    preferredDuringSchedulingIgnoredDuringExecution:
+      - weight: 100
+        podAffinityTerm:
+          labelSelector:
+            matchExpressions:
+              - key: app.kubernetes.io/name
+                operator: In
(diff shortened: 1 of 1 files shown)


charts/vector/values.yaml
```diff
@@ -178,7 +178,19 @@ tolerations: []
...
+  podAntiAffinity:
+    preferredDuringSchedulingIgnoredDuringExecution:
+      - weight: 100
+        podAffinityTerm:
+          labelSelector:
+            matchExpressions:
+              - key: app.kubernetes.io/name
+                operator: In
+                values:
+                  - vector
```
spencergilbert, line 192: This would not be ideal depending on our story around installing Vector as both an Agent and Aggregator in the same cluster.

silversupreme:
Thinking there might be a use instead for like a `scaleAcrossZones` option that can be enabled, and ignored if you set a custom `affinity`. We can default that to `false` but make it easy to change.

I'll try to work that in instead, will put up a diff momentarily.

spencergilbert:
> Thinking there might be a use instead for like a `scaleAcrossZones` option that can be enabled, and ignored if you set a custom `affinity`. We can default that to `false` but make it easy to change.
> 
> I'll try to work that in instead, will put up a diff momentarily.

I definitely don't mind having the `preferred` config by default since it doesn't actively block deployments ever, I just want to make sure we're all aligned on where the chart is heading and advertising that appropriately to users

PR #176, opened 2022-03-16
docs(vector): Add quickstart values and docs

spencergilbert:

~~I'd like to have these in the `ci/` folder as well, but symlinking them failed - happy to take suggestions~~ 😆 

Added a step to `cp` the values files from `examples/` into `ci/`


charts/vector/docs/datadog_quickstart.yaml
```diff
@@ -0,0 +1,198 @@
...
+  transforms:
+    remap_logs:
+      type: remap
+      inputs:
+        - datadog_agent.logs
+        - internal_logs
+      source: |
+        # TODO: Having Datadog Agents collect Aggregator logs would provide better enrichment,
+        # internal_logs will also provide the Pod's name as the host facet in Datadog rather
+        # than the name of the Node
```
spencergilbert, line 90: It's probably a better experience on the Datadog side to remove the `console` sink and rely on `vector tap`.

charts/vector/docs/eks_quickstart.yaml
```diff
@@ -0,0 +1,21 @@
+## See Vector helm documentation to learn more:
+## https://vector.dev/docs/setup/installation/package-managers/helm/
+
+## Configuration for Vector's Service
+service:
+  # service.type -- Set type of Vector's Service
+  type: "LoadBalancer"
```
spencergilbert, line 7: As I understand it, this will _always_ create an NLB, and we'd need to use an `Ingress` to provision an ALB. From what I can tell (from [here](https://aws.amazon.com/elasticloadbalancing/features/)):

```
ALB can handle: HTTP/S, gRPC
NLB can handle: TCP, UDP, TLS
```

We likely need some more hands-on testing/knowledge on what we'd prefer and the limitations of each. The `eks_quickstart.yaml` could also be split into a `eks_alb|nlb_quickstart.yaml` as well.

charts/vector/docs/quickstart.md
```diff
@@ -0,0 +1,168 @@
...
+    logs_config:
+      logs_dd_url: vector-haproxy.vector:8282
+      logs_no_ssl: true # Set to false if SSL is enabled in Vector source configuration
+      use_http: true
+VALUES
+```
+
+**WARNING:** The above config disables TLS verification of the connection to the
+Kubelet. This is useful in a quickstart environment like `minikube`, but when
+real certificates are available, the `kubelet_tls_verify` option should be removed.
```
spencergilbert, line 118: I'm not super happy with the phrasing here, just want to _not_ recommend turning off security features in prod

charts/vector/docs/quickstart.md
```diff
@@ -0,0 +1,168 @@
...
+
+_NOTE: When you have the `datadog_agent` source configured, a `helm install` or
+`helm upgrade` will output the appropriate values to provide to the datadog helm
+chart._
```
spencergilbert, line 122: This feature is coming in https://github.com/vectordotdev/helm-charts/pull/178 - writing and testing the config for Datadog Agents here was done first to get the values right

charts/vector/docs/quickstart.md
```diff
@@ -0,0 +1,168 @@
...
+
+## Moving to the cloud
+
+### Elastic Kubernetes Service
+
+#### Prerequisites
+
+#### Installing
+
+#### Troubleshooting
```
spencergilbert, line 168: Might link out to separate docs for each cloud provider rather than have everything in one file...

spencergilbert:
Discussing with @silversupreme I'll be splitting out the AWS work to get Datadog landed faster and let us do some more internal testing with AWS resources.

charts/vector/docs/quickstart.md
silversupreme: I think it might be better to format the above snippet as:

```
agents:
  useConfigMap: true
  customAgentConfig:
    # NOTE: If you're using a quickstart environment like `minikube`, include the line below; otherwise, we recommend leaving it with the default setting of `true`.
    # kubelet_tls_verify: false
```

Abstractly, I think it's good to treat things like `minikube` as deviations, and make recommendations prod-ready by default.

charts/vector/docs/quickstart.md
```diff
@@ -0,0 +1,168 @@
...
+`helm upgrade` will output the appropriate values to provide to the datadog helm
+chart._
+
+```bash
+helm install datadog datadog/datadog --namespace datadog --create-namespace \
+	--values dd-agent.yaml --set datadog.apiKey="${DATADOG_API_KEY}"
+```
+
+Once the Datadog Agents are running you should be able to see your logs in
+Datadog's [Live Tail](https://app.datadoghq.com/logs/livetail?query=sender%3Avector) view,
```
silversupreme, line 130: Nice link, that's a smart idea.

charts/vector/docs/quickstart.md
spencergilbert: Sorta feel like making prod-ready rec's in a "quickstart" guide seems backwards 🤔

spencergilbert:
@silversupreme looking at this again - I think given the premise, it's fair to say the datadog agent will always be included for _this_ quickstart since it's part of the guide itself - because of that I didn't add in any guidance around sending `internal_logs` when there isn't a separate collector agent.

silversupreme (approved):
340vrl

This matches basically exactly what worked for the POV I did this week, and will be a great help for folks. Nice work.

Issue #177, opened 2022-03-17
Output Datadog Helm Chart values on install


silversupreme:
When we setup a customer using the `datadog_agent` source, it would be extremely helpful to have the chart emit the necessary configuration to properly wire the agent up to the new Aggregator.

This would need to be updated once the `vector` config options land in the Agent, but should suffice for now:

```yaml
datadog:
  # These need to be enabled for the Agent to even collect logs at all.
  logs:
    enabled: true
    containerCollectAll: true

  # Don't collect Vector logs so we can avoid a logging loop.
  containerExclude: "image:vector"

  # Override the Logs Agent to point to Vector instead.
  env:
    - name: DD_LOGS_CONFIG_DD_URL
      value: "${URL}"
    - name: DD_LOGS_CONFIG_USE_HTTP
      value: "true"
```

spencergilbert:
Think the `env` based settings are better/easier than the [`YAML`](https://github.com/vectordotdev/vector-demos/blob/main/eks-aggregator/helm/datadog.yaml#L13-L20)?

silversupreme:
I think that's entirely fair. I'll amend the description accordingly.

spencergilbert:
Going through the quickstart work https://github.com/vectordotdev/helm-charts/pull/176, I think we'll need to provide it more like:
```yaml
datadog:
  containerExclude: "name:vector"
  logs:
    enabled: true
    containerCollectAll: true
agents:
  useConfigMap: true
  customAgentConfig:
    kubelet_tls_verify: false
    logs_config:
      logs_dd_url: {URL}
      logs_no_ssl: true
      use_http: true
```

silversupreme:
I had a feeling that the Helm values would be necessary. Sad trombone. Thanks for verifying though.

PR #178, opened 2022-03-18
feat(vector): Add help output for configurations with datadog_agent source

spencergilbert:

Closes: https://github.com/vectordotdev/helm-charts/issues/177

NOTE: I've varied from the issue on the setting of `DD_CONTAINER_EXCLUDE` to use `name` instead of `image` to more gracefully handle users bringing their own images.

- [x] Detect when a `datadog_source` is configured
- [x] Output SSL config based on `tls` configuration
- [x] Parse `address` to get the port and build the url from that and the service name
- [ ] Stretch goal - get the load balancer/ingress name for the url [if possible](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-lookup-function)
- [ ] Handle multiple configured `datadog_source` components

I'd like to push the last two into separate issues - duplicate sources doesn't break the current implementation, it just prints out the values for one. The loadbalancer improvement will take some time with a real cluster


silversupreme (approved):
Love it. Might be worthwhile to swap the versions while the new style is pre-release but I'm not gonna block on that. Nice work.

charts/vector/templates/_helpers.tpl
```diff
@@ -241,9 +241,95 @@ Print the `url` option for the Vector command.
   {{- if $.Values.customConfig -}}
     {{- if $.Values.customConfig.api -}}
       {{- if $.Values.customConfig.api.address -}}
---url {{ printf "http://%s/graphql" $.Values.customConfig.api.address }}
+{{ print "\\" }}
```
StephenWakely, line 244: Should this line be indented?

charts/vector/templates/_helpers.tpl
```diff
@@ -241,9 +241,95 @@ Print the `url` option for the Vector command.
...
+      {{ include "_fmt.blue" "logs:" }}
+        {{ include "_fmt.blue" "enabled:" }} {{ include "_fmt.yellow" "true" }}
+        {{- if not .Values.haproxy.enabled }}
+        {{ include "_fmt.blue" "url:" }} "{{ $protocol }}://{{ include "vector.fullname" $ }}.{{ $.Release.Namespace }}:{{ $sourceDatadogAgentPort | default "8282" }}"
+        {{- else }}
+        {{ include "_fmt.blue" "url:" }} "{{ $protocol }}://{{ include "haproxy.fullname" $ }}.{{ $.Release.Namespace }}:{{ $sourceDatadogAgentPort | default "8282" }}"
+        {{- end }}
+      {{ include "_fmt.blue" "metrics:" }}
+        {{ include "_fmt.blue" "enabled:" }} {{ include "_fmt.yellow" "true" }}
+        {{- if not .Values.haproxy.enabled }}
```
StephenWakely, line 305: I think it would be clearer to not have the not and swap the if and else blocks around..

```suggestion
        {{- if .Values.haproxy.enabled }}
```

StephenWakely (approved):
Looks really good! Just some very minor points...

charts/vector/templates/_helpers.tpl
spencergilbert: No, it's getting used at the end of the command [here](https://github.com/vectordotdev/helm-charts/blob/708069a986676afa244736accaea896e9c5b0874/charts/vector/templates/_helpers.tpl#L198), to provide some line-wrapping functionality since appending `--url http://127.0.0.1:8686/graphql` or what not gets pretty far into the right half of the screen.

charts/vector/templates/_helpers.tpl
```diff
@@ -241,9 +241,95 @@ Print the `url` option for the Vector command.
...
+  {{ include "_fmt.blue" "logs:" }}
+    {{ include "_fmt.blue" "enabled:" }} {{ include "_fmt.yellow" "true" }}
+    {{ include "_fmt.blue" "containerCollectAll:" }} {{ include "_fmt.yellow" "true" }}
+{{ include "_fmt.blue" "agents:" }}
+  {{ include "_fmt.blue" "useConfigMap:" }} {{ include "_fmt.yellow" "true" }}
+  {{ include "_fmt.blue" "customAgentConfig:" }}
+    {{ include "_fmt.blue" "kubelet_tls_verify:" }} {{ include "_fmt.yellow" "false" }}
+    {{ include "_fmt.blue" "logs_config:" }}
+      {{- if not .Values.haproxy.enabled }}
+      {{ include "_fmt.blue" "logs_dd_url:" }} "{{ include "vector.fullname" $ }}.{{ $.Release.Namespace }}:{{ $sourceDatadogAgentPort | default "8282" }}"
```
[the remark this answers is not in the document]
spencergilbert: I don't believe so, all of these options should be at the same indention depth - _honestly_ we should be using `nindent` or `indent` to provide whitespace via the templating engine, rather than my including literal spaces at the beginning of these lines. However the templating engine was getting angry and I got tired of fighting with it, so I settled for this easier but more brittle/less idiomatic way.

If we do a second pass on this to clean up the templates and make things more re-usable functions we should be able to use the indention functions more easily.

PR #179, opened 2022-03-21
fix(vector): update app.kubernetes.io/version to be the image.tag if it is set.

### LINKED ISSUE #174: Label app.kubernetes.io/version isn't updated with image tag
app.kubernetes.io/version label in [helper template](https://github.com/vectordotdev/helm-charts/blob/87ff789d7f58dd434d0197deccb957c040624650/charts/vector/templates/_helpers.tpl#L39) isn't updated with image tag like this "{{ .Values.image.tag | default .Chart.AppVersion }}" . This results in incorrect label with image tag is non-default.

StephenWakely:
Closes #174 

This updates the `app.kubernetes.io/version` field to be `.Values.image.tag` if it has been set.

I have made this change everywhere, including in the haproxy templates. Let me know if that is too much!

Signed-off-by: Stephen Wakely <fungus.humungus@gmail.com>

### DIFF
--- a/charts/vector-shared/templates/_labels.tpl
+++ b/charts/vector-shared/templates/_labels.tpl
@@ -7,7 +7,7 @@ Common labels.
 helm.sh/chart: {{ include "libvector.chart" . }}
 {{ include "libvector.selectorLabels" . }}
 {{- if .Chart.AppVersion }}
-app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+app.kubernetes.io/version: {{ .Values.image.tag | default .Chart.AppVersion }}
 {{- end }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 app.kubernetes.io/part-of: {{ include "libvector.name" . }}
--- a/charts/vector/templates/haproxy/_helpers.tpl
+++ b/charts/vector/templates/haproxy/_helpers.tpl
@@ -14,7 +14,7 @@ Common labels
 helm.sh/chart: {{ include "vector.chart" . }}
 {{ include "haproxy.selectorLabels" . }}
 {{- if .Chart.AppVersion }}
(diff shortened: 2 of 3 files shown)


charts/vector-shared/templates/_labels.tpl
```diff
@@ -7,7 +7,7 @@ Common labels.
 helm.sh/chart: {{ include "libvector.chart" . }}
 {{ include "libvector.selectorLabels" . }}
 {{- if .Chart.AppVersion }}
-app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+app.kubernetes.io/version: {{ .Values.image.tag | default .Chart.AppVersion }}
```
spencergilbert, line 10: I'd leave this as-is since it's deprecated

charts/vector/templates/haproxy/_helpers.tpl
```diff
@@ -14,7 +14,7 @@ Common labels
 helm.sh/chart: {{ include "vector.chart" . }}
 {{ include "haproxy.selectorLabels" . }}
 {{- if .Chart.AppVersion }}
-app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+app.kubernetes.io/version: {{ .Values.image.tag | default .Chart.AppVersion }}
```
spencergilbert, line 17: It's odd that this isn't the HAProxy version... but perhaps I intentionally did that for a reason I no longer remember. What do you think?

spencergilbert (changes requested):

spencergilbert:
Sorry @StephenWakely I'm a dummy and didn't notice the `| quote` got removed - can we still pipe that in to ensure people can't pass bad values into the label's value? Kubernetes rejects if it's not a string

spencergilbert (approved):

PR #203, opened 2022-04-26
fix(vector): no need to set replicas to statefulset  when using hpa

CasparChou:
as title,
deploymentset has an if condition but statefulset doesn't have, but I think this is the same thing when using autoscaling

### DIFF
--- a/.github/cliff.toml
+++ b/.github/cliff.toml
@@ -52,3 +52,4 @@ commit_parsers = [
 filter_commits = true
 # glob pattern for matching git tags
 tag_pattern = "vector-*"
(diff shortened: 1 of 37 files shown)


spencergilbert (approved):
Good catch - thanks! Definitely a forgotten inclusion when we added HPA support to Statefulsets.

CasparChou:
sorry for missing the chart version
`helm-docs` seems no change in this 

Should I modify CHANGELOG.md  and README.md?

spencergilbert:
Looks perfect, thanks! I'll look to get this published shortly.

PR #211, opened 2022-05-20
feat(vector): add RBAC to list and watch nodes

nabokihms:
The ability to list and watch nodes will be required to annotate logs from `kubernetes_logs` sources in the future. 

See https://github.com/vectordotdev/vector/pull/12730 for more details.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,14 @@ All notable changes to this project will be documented in this file.
 
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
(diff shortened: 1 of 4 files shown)


spencergilbert:
As a nit - I think we should bump the minor version and have this as a feat, as it's adding new functionality.

Thanks for the update!

spencergilbert:
More nitpicking 😓 could you edit commit as well so it makes it into the changelog as a feat rather than a fix?

nabokihms:
Sorry, my bad. It should be ok now.

spencergilbert (approved):
Thanks so much! I'll get this published today

PR #219, opened 2022-06-13
feat(vector): Add digest option to vector image

tomer-epstein:
Signed-off-by: Tomer Epstein <tomer.epstein@sap.com>

**Motivation**
add better integrity to the image version, we should provide a way to add a digest option to the chart images this way we can validate the image version

**Approach**
How does this change address the problem?

Add digest option to vector chart,
By adding a new sha field to the pod file and concatenate it to the image string address when provided.

**Pull Request status**
- [x]  Initial implementation
- [ ]  Refactoring
- [ ]  Unit tests
- [ ] Integration tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,14 @@ All notable changes to this project will be documented in this file.
 
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
+## [vector-0.13.0] - 2022-06-13
+
+### Vector
+
+#### Features
+
+- Add digest option to vector image ([6e54a0a](https://github.com/vectordotdev/helm-charts/commit/6e54a0a463c47ec57e5f8b315e40a07ae584b878))
+
 ## [vector-0.12.0] - 2022-06-01
 
 ### Vector
--- a/charts/vector/Chart.yaml
+++ b/charts/vector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: vector
-version: "0.12.1"
+version: "0.13.0"
 kubeVersion: ">=1.15.0-0"
 description: A lightweight, ultra-fast tool for building observability pipelines
 type: application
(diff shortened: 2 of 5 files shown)


spencergilbert:
Thanks for the contribution - Digests would be usable today by just passing `image.tag: 0.22.1-distroless-libc@sha256:ba208391be6f7705f5e7bf7513ad7600cf8cc2c5dff3a4691632de3afbf0bcd0`, right? This is just a more friendly way of exposing it?

tomer-epstein:
It's more then just  a friendly way.

by passing a tag that conains a digest, the labels section in the vector chart will fail.
https://github.tools.sap/api-gateway/vector-chart-wrapper/blob/65a14f1ad36255eb3146f8df826b6221cd617d29/vector/charts/vector/templates/_helpers.tpl#L40

label values in k&s are allowing only specific syntax.
https://github.tools.sap/api-gateway/vector-chart-wrapper/blob/65a14f1ad36255eb3146f8df826b6221cd617d29/vector/charts/vector/templates/_helpers.tpl#L40

Example of using digest (and tag as comment before the digest) OR tag in grafana:
https://github.com/grafana/helm-charts/blob/main/charts/grafana/templates/_pod.tpl#L397-L401

spencergilbert:
> It's more then just a friendly way.
> 
> by passing a tag that conains a digest, the labels section in the vector chart will fail. https://github.tools.sap/api-gateway/vector-chart-wrapper/blob/65a14f1ad36255eb3146f8df826b6221cd617d29/vector/charts/vector/templates/_helpers.tpl#L40
> 
> label values in k&s are allowing only specific syntax. https://github.tools.sap/api-gateway/vector-chart-wrapper/blob/65a14f1ad36255eb3146f8df826b6221cd617d29/vector/charts/vector/templates/_helpers.tpl#L40
> 
> Example of using digest (and tag as comment before the digest) OR tag in grafana: https://github.com/grafana/helm-charts/blob/main/charts/grafana/templates/_pod.tpl#L397-L401

Ah good point - I didn't think about that. I haven't written up a contributing doc yet, but we'll need to bump the minor version and run `helm-docs` to update the readme. I'm happy to do that if you'd prefer.

tomer-epstein:
Sure, go for it.
Will you do it on this PR?

spencergilbert (approved):
🤦

tomer-epstein:
> 🤦
Tnx @spencergilbert 
Appriciate  the very quick response.

PR #221, opened 2022-06-15
feat(vector): Add option to use KEDA autoscaler

artmakh:
Motivation
[KEDA](https://keda.sh/) provides possibility to scale pods based on any metric (or query) available in prometheus or [many overs](https://keda.sh/docs/2.7/scalers/) datasources. Since vector have wide variety of internal metrics better suitable for scaling decisions, it's nice to have possibility to choose KEDA instead of plain HPA

Approach
How does this change address the problem?

It adds necessary values to helm chart and KEDA ScaledObject template

### DIFF
--- a/charts/vector/templates/deployment.yaml
+++ b/charts/vector/templates/deployment.yaml
@@ -6,7 +6,7 @@ metadata:
   labels:
     {{- include "vector.labels" . | nindent 4 }}
 spec:
-  {{- if not .Values.autoscaling.enabled }}
+  {{- if and (not .Values.autoscaling.enabled) (not .Values.keda.enabled) }}
   replicas: {{ .Values.replicas }}
   {{- end }}
   selector:
--- a/charts/vector/templates/keda.yaml
+++ b/charts/vector/templates/keda.yaml
@@ -1,0 +1,34 @@
+{{- if or (eq .Values.role "Aggregator") (eq .Values.role "Stateless-Aggregator") -}}
+{{- if and .Values.keda.enabled (not .Values.autoscaling.enabled)}}
+apiVersion: keda.sh/v1alpha1
+kind: ScaledObject
+metadata:
+  name: {{ include "vector.fullname" . }}
+  labels:
+    {{- include "vector.labels" . | nindent 4 }}
+  {{- if .Values.keda.scaledObject.annotations }}
+  annotations: {{ toYaml .Values.keda.scaledObject.annotations | nindent 4 }}
+  {{- end }}
+spec:
+  scaleTargetRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: {{ include "vector.fullname" . }}
+  pollingInterval: {{ .Values.keda.pollingInterval }}
+  cooldownPeriod: {{ .Values.keda.cooldownPeriod }}
+  minReplicaCount: {{ .Values.keda.minReplicas }}
+  maxReplicaCount: {{ .Values.keda.maxReplicas }}
+  triggers:
+{{- with .Values.keda.triggers }}
+{{ toYaml . | indent 2 }}
+{{ end }}
+{{- if .Values.keda.behavior }}
+  advanced:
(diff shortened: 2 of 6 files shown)


spencergilbert:
Thanks for the contribution! I'll try and get a review on this early next week.

charts/vector/templates/deployment.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
   labels:
     {{- include "vector.labels" . | nindent 4 }}
 spec:
-  {{- if not .Values.autoscaling.enabled }}
+  {{- if and (not .Values.autoscaling.enabled) (not .Values.keda.enabled) }}
```
spencergilbert, line 9: I think this should be an `or`? If either of them is enabled we don't want to include the replicas key.

charts/vector/templates/statefulset.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
   labels:
     {{- include "vector.labels" . | nindent 4 }}
 spec:
-  {{- if not .Values.autoscaling.enabled }}
+  {{- if and (not .Values.autoscaling.enabled) (not .Values.keda.enabled) }}
```
spencergilbert, line 9: Same as above: I think this should be an `or`? If either of them is enabled we don't want to include the replicas key.

charts/vector/templates/keda.yaml
```diff
@@ -0,0 +1,34 @@
+{{- if or (eq .Values.role "Aggregator") (eq .Values.role "Stateless-Aggregator") -}}
+{{- if and .Values.keda.enabled (not .Values.autoscaling.enabled)}}
```
spencergilbert, line 2: What do you think of raising a more explicit error if both are configured, like I've done here with [`fail`](https://github.com/vectordotdev/helm-charts/blob/develop/charts/vector/templates/_pod.tpl#L120)

spencergilbert, line 2:
```suggestion
{{- if and .Values.keda.enabled (.Capabilities.APIVersions.Has "keda.sh/v1alpha1") (not .Values.autoscaling.enabled)}}
```

Does this work as well? I don't know if I've tried this with custom resources, but I assume it should work.

spencergilbert (commented):
Sorry for the long delay, I got behind on a lot of things 😓

charts/vector/templates/deployment.yaml
artmakh: As far as i understand `or` won't work here. I found that under the hood helm [uses go template lib](https://pkg.go.dev/text/template#hdr-Functions) , as stated [here](https://github.com/helm/helm/issues/6250#issuecomment-523134350). So `and` should be read as `if x then y else x.` . Here is small example how it works - https://go.dev/play/p/kDiNqowjId4 . I've also checked it with `helm install --dry-run` and verified that there is no replicas in either way.

charts/vector/templates/keda.yaml
artmakh: Good point, added

spencergilbert:
Sorry - this is still on my radar, just been trying to finish up tasks from the last quarter.

spencergilbert (commented):
Sorry for the long delay - I'm having a little trouble getting this to render locally. I installed KEDA via

```shell
kubectl apply -f https://github.com/kedacore/keda/releases/download/v2.7.1/keda-2.7.1.yaml
```

and confirmed the api-version matches the cap we're looking for:

```shell
❯ kubectl api-versions | rg keda
keda.sh/v1alpha1
```

But I'm not getting the output rendered when I enable keda. I do get the error message when I enabled both though:

```shell
❯ helm template vector . --set autoscaling.enabled=true --set keda.enabled=true
Error: execution error at (vector/templates/keda.yaml:2:63): You can't enable both KEDA and HPA, please choose only one
```

spencergilbert:
Closing due to inactivity

PR #223, opened 2022-06-27
Fix HAProxy Configmap Conditional

jonwinton:
# Purpose

The `ConfigMap` template is checking for the `customConfig` property for Vector as opposed to the HAProxy `customConfig`. If using `customConfig` for Vector, the rendered HAProxy config is empty and Kubernetes probes fail.

### DIFF
--- a/charts/vector/templates/haproxy/configmap.yaml
+++ b/charts/vector/templates/haproxy/configmap.yaml
@@ -7,7 +7,7 @@ metadata:
     {{- include "haproxy.labels" . | nindent 4 }}
 data:
   haproxy.cfg: |
-  {{- if .Values.customConfig }}
+  {{- if .Values.haproxy.customConfig }}
   {{ tpl .Values.haproxy.customConfig . | nindent 4 }}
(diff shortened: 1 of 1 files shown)


charts/vector/templates/haproxy/configmap.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
     {{- include "haproxy.labels" . | nindent 4 }}
 data:
   haproxy.cfg: |
-  {{- if .Values.customConfig }}
+  {{- if .Values.haproxy.customConfig }}
```
spencergilbert, line 10: Looking at this I think we actually need a bit more of a robust checking. If Vector is using a non-default config, the default config provided for HAProxy isn't guaranteed to route correctly - and the reverse is also true.

Perhaps checking if either are custom and raising if that's the case would be helpful - what do you think?

spencergilbert (commented):
Thanks, this does seem like a typo in the original implementation - but I'm not sure the fix fully covers the UX.

PR #230, opened 2022-07-19
feat(vector): Add lifecycle hooks

tombokombo:
Hi, there are missing optional lifecycle hooks, so this PR reintroduce them, as they are in deprecated vector-agent chart.


spencergilbert:
Thanks for the contribution @tombokombo!

spencergilbert (approved):

PR #232, opened 2022-08-02
Add support for autoscaling v2

tuananhnguyen-ct:
Signed-off-by: Tuan Anh Nguyen <tuananh.nguyen-ext@commercetools.de>

Fixes https://github.com/vectordotdev/helm-charts/issues/186

### DIFF
--- a/charts/vector/Chart.yaml
+++ b/charts/vector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: vector
-version: "0.14.0"
+version: "0.14.1"
 kubeVersion: ">=1.15.0-0"
 description: A lightweight, ultra-fast tool for building observability pipelines
 type: application
--- a/charts/vector/Chart.yaml
+++ b/charts/vector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: vector
-version: "0.14.1"
+version: "0.15.0"
 kubeVersion: ">=1.15.0-0"
 description: A lightweight, ultra-fast tool for building observability pipelines
 type: application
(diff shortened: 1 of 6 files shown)


charts/vector/templates/haproxy/_helpers.tpl
```diff
@@ -38,3 +38,14 @@ Create the name of the service account to use
...
 {{- end }}
+
+{{/*
+Return the appropriate apiVersion for HPA autoscaling APIs.
+*/}}
+{{- define "autoscaling.apiVersion" -}}
+{{- if or (.Capabilities.APIVersions.Has "autoscaling/v2") (semverCompare ">=1.23" .Capabilities.KubeVersion.Version) -}}
+"autoscaling/v2"
+{{- else -}}
+"autoscaling/v2beta2"
```
spencergilbert, line 49: I think we'll want to keep a check for `v2beta2` support to not cause issues with clusters that are so old that they don't even support that version. Though I'm not seeing how far back that would end up being.

charts/vector/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: vector
-version: "0.14.0"
+version: "0.14.1"
```
spencergilbert, line 3: I'd classify this as a new feature/enhancement and bump it to `0.15`

spencergilbert (changes requested):

charts/vector/templates/haproxy/_helpers.tpl
tuananhnguyen-ct: v2beta2 is available from k8s 1.12, and it'll require some changes in HPA config (targetAverageUtilization vs averageUtilization) if we go back to v2beta1, while vector only support k8s 1.15+. I don't see why we really need to go back for v2beta1 support for now.

spencergilbert: Thanks - that's definitely far back enough!

spencergilbert (approved):

PR #235, opened 2022-08-10
fix(vector): Actually release version bump

jszwedko:
Accidentally merged in develop to master without the commit bumping the
Vector version.

Signed-off-by: Jesse Szwedko <jesse@szwedko.me>


spencergilbert (approved):
😬 exactly as we intended...

PR #238, opened 2022-08-13
feat(vector): Add support for setting the pod hostNetwork (#213)

### LINKED ISSUE #213: vector _pod.tpl does not allow hostNetwork
For clusters with `--http-put-response-hop-limit 1` on purpose, we need to run vector with `hostNetworking: true` so it can run the `aws_ec2_metadata` transform without exposing IMDSv2 to the workload Pods

The [current `_pod.tpl`](https://github.com/vectordotdev/helm-charts/blob/develop/charts/vector/templates/_pod.tpl) does not offer `.Values.hostNetwork` or such that would allow the chart to be installed in that configuration

rgl:
This closes #213

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,14 @@ All notable changes to this project will be documented in this file.
 
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
+## [vector-0.16.0] - 2022-08-15
+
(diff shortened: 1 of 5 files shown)


spencergilbert:
Thanks for the contribution, we'll look to cut a new version soon!

spencergilbert (approved):

PR #243, opened 2022-09-28
fix: extraVolumeMounts

AvihaiSam:
`extraVolumeMounts` should be located before `extraContainers`
otherwise, if extraContainers is not null, yaml is not valid and extraVolumeMounts is impossible to use.


spencergilbert:
Thanks for catching that @AvihaiSam! I've included an additional test in CI to ensure this doesn't regress.

PR #252, opened 2022-10-11
fix(vector): :bug: pod tpl

### LINKED ISSUE #251: Release 0.16.2 issue with env
This commit:
https://github.com/vectordotdev/helm-charts/commit/89be2e2a77e7f11c57c829c6c49809bc3ab26eda
introduced an issue where if we don't have ".Values.env" defined, the very important useful environment variables VECTOR_SELF_NODE_NAME (and others) are skipped, as they are inside the "if".

Could this be fixed ? 
(just needs to move the "{{ -end }}" a few lines above).

Thanks

CC @dannylongeuay and @spencergilbert

dannylongeuay:
Fixes #251

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,14 @@ All notable changes to this project will be documented in this file.
 
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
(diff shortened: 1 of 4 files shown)


spencergilbert (approved):
😮‍💨 I didn't catch that either, definitely room for improvement on CI here as well.

PR #256, opened 2022-10-19
fix(vector): Apply various suggested changes from JetBrains markdown and proofreading inspections

spencergilbert:

### DIFF
--- a/charts/vector/Chart.yaml
+++ b/charts/vector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: vector
-version: "0.16.3"
+version: "0.16.4"
 kubeVersion: ">=1.15.0-0"
 description: A lightweight, ultra-fast tool for building observability pipelines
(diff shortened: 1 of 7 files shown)


charts/vector/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: vector
-version: "0.16.3"
+version: "0.16.4"
```
StephenWakely, line 3: Why is this necessary?

StephenWakely: Ah.. has this leaked in from the colour codes pr?

spencergilbert: We'd fail the check [here](https://github.com/vectordotdev/helm-charts/actions/runs/3282096145/jobs/5405027736#step:5:29).

The datadog agent does similarly, https://github.com/DataDog/helm-charts/pull/681

PR #257, opened 2022-10-19
feat(vector): Support configuration for both the normal and headless service

### LINKED ISSUE #253: Separate configuration for the Service and Headless Service
It's been reported that when configuring the chart with [aws-load-balancer-controller annotations](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/guide/service/annotations/#lb-source-ranges) the fact that both the Service and Headless Service are configured with the same values is causing problems.

The chart should be enhanced to allow for granular control of the Services, if required.

EDIT: headless services are hardcoded as ClusterIPs, using the aws-lb the type needs to be LoadBalancer

spencergilbert:
The headless service can be disabled and configured separately, but the default behavior will be to reuse the config from the service.

Closes #253


charts/vector/ci/serviceHeadless-disabled.yaml
```diff
@@ -0,0 +1,8 @@
+# Separately disable the Headless Service
+## Values file for testing separate enabling for service and serviceHeadless.
```
neuronull, line 2: I'm not super familiar with yaml comments, is there a difference between one # and two ## ?

charts/vector/templates/service-headless.yaml
```diff
@@ -1,4 +1,63 @@
-{{- if .Values.service.enabled }}
+{{/*
+TODO: Remove outer "if/elseif"
```
neuronull, line 2: Still a valid TODO ?

neuronull (approved):

charts/vector/ci/serviceHeadless-disabled.yaml
spencergilbert: No, both are comments... I might have copied it from somewhere. 😅 It's been a while, I think it was also a title => description sort of relationship...

charts/vector/templates/service-headless.yaml
spencergilbert: Yeah, we should be able to have just a single `{{- if .Values.serviceHeadless.enabled }}` wrapping the entire resource, just to clean things up. The existing if/elif is needed because the upgrade from the previous version to this version would fail. We can remove it for the upgrade from this version to the next (theoretically, depends on how much grace we want to provide for upgrades).

charts/vector/README.md
```diff
@@ -135,13 +135,13 @@ helm install --name <RELEASE_NAME> \
 | command | list | `[]` | Override Vector's default command |
 | commonLabels | object | `{}` | Add additional labels to all created resources |
 | containerPorts | list | `[]` | Manually define Vector's Container ports, overrides automated generation of Container ports |
-| customConfig | object | `{}` | Override Vector's default configs, if used **all** options need to be specified # This section supports using helm templates to populate dynamic values # Ref: https://vector.dev/docs/reference/configuration/ |
```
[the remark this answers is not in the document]
spencergilbert: Yeah, an older version of `helm-docs` used to not render comments that were `##` - now it just renders the second `#` in the output. It was going to be part of the general cleanup I decided to not pile into this PR 👍

charts/vector/ci/serviceHeadless-disabled.yaml
spencergilbert: Datadog's charts have just the `# title`, maybe at some point I felt like one needed more explanation and then I did that 🤷

charts/vector/templates/service-headless.yaml
neuronull: Ah, makes sense.

PR #261, opened 2022-10-24
Add Pod topology spread constraints chart value

### LINKED ISSUE #260: Add support for Pod topology spread constraints
The `vector` Helm chart currently doesn't have support for [Pod topology spread constraints](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/). Adding the topology spread constraints will help with increasing the fault tolerance of Vector in [the aggregator role](https://vector.dev/docs/setup/deployment/roles/#aggregator) running as a Deployment or StatefulSet setups by spreading the Pods evenly over failure domains (e.g. nodes, zones).

My specific use-case is the Deployment variant with HorizontalPodAutoscaler defined. The usage of autoscaling makes static Pod allocations cumbersome.

This doesn't make sense for the DaemonSet agent role setups of course.

fkrestan:
Resolves #260


charts/vector/ci/aggregator-all-values.yaml
```diff
@@ -41,6 +41,16 @@ affinity:
           - e2e-az1
           - e2e-az2
 
+topologySpreadConstraints:
+  - labelSelector:
+      matchLabels:
+        app.kubernetes.io/name: vector
+        app.kubernetes.io/instance: vector
+        app.kubernetes.io/component: Aggregator
```
fkrestan, line 49: The use-case where the `topologySpreadConstraints.labelSelector` is the same as the label selectors used in other related API objects selecting the workload Pods is very common. It would be nice to be able to just say `{{ include "vector.selectorLabels" . }}` in the user-provided `values.yaml`.

I've tried tinkering around with the Helm [tpl function](https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-tpl-function), but the results were not very ergonomic.

fkrestan:
> I merged some changes that will probably need to be merged in/resolved against your changes.
I've re-based against the latest `develop` and updated the docstring to match the new style.

spencergilbert:
Great, thank you! I'm aiming to play around locally and get this reviewed/merged by the end of the week.

charts/vector/values.yaml
```diff
@@ -206,6 +206,10 @@ tolerations: []
 # rules for Vector Pods.
 affinity: {}
 
+# topologySpreadConstraints -- Configure [topology spread constraints](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/)
+# for Vector Pods.
```
spencergilbert, line 210:
```suggestion
# for Vector Pods. Valid for the "Aggregator" and "Stateless-Aggregator" roles.
```

charts/vector/templates/_pod.tpl
```diff
@@ -172,6 +172,10 @@ affinity:
 tolerations:
 {{ toYaml . | indent 2 }}
 {{- end }}
+{{- with  .Values.topologySpreadConstraints }}
+topologySpreadConstraints:
+{{- toYaml . | nindent 2 }}
+{{- end }}
```
spencergilbert, line 178:
```suggestion
{{- if  .Values.topologySpreadConstraints }}
topologySpreadConstraints:
  labelSelector:
    matchLabels:
{{ include "vector.selectorLabels" . | nindent 6 }}
{{- toYaml .Values.topologySpreadConstraints | nindent 2 }}
{{- end }}
```

Not perfect, but this allows you to just pass in key/values of the constraints and re-use the selectorLabels.

charts/vector/values.yaml
spencergilbert: Added in https://github.com/vectordotdev/helm-charts/pull/261/commits/505056d25a8d19a2956df9ccd34d627197a8a45d and regen'd the docs.

spencergilbert:
@fkrestan do we need to check that this hasn't been set if the user is deploying a Daemonset? Would that error, or just be ignored by the API?

fkrestan:
> @fkrestan do we need to check that this hasn't been set if the user is deploying a Daemonset? Would that error, or just be ignored by the API?

I've tested this and it seems to be ignored at least in effect:
- The spread constraint gets applied and doesn't get stripped form the DaemonSet spec. `topologySpreadConstraints` is part of the Pod API spec which is embedded in the DaemonSet so it makes sense that this passes API validation.
- Looking at the `kubectl describe` output, it seems to be applied and respected. I don't see any complaints in Kubernetes API nor controller manager logs.
- Looking at the [documentation](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/#topologyspreadconstraints-field), it doesn't seem possible to create a constraint that would not get satisfiedg by DaemonSet.

IMO there is no need to put any checks in place.

spencergilbert (approved):
Awesome, thanks for double checking - I expect I'll cut the release this week.

PR #265, opened 2022-11-03
fix(vector): Update haproxy to use image.pullSecrets

SIPR-octo:
Very simple fix, the template for haproxy deployment seems to use the old naming for imagePullSecrets

Thanks in advance! :smile_cat:


spencergilbert (approved):
Doh, good catch!

PR #268, opened 2022-11-09
fix HorizontalPodAutoscaler helm template api version

AvihaiSam:
# Description
ArgoCD uses `helm template` which fails to get the right api-version on k8s-version 1.21
On 1.21 cluster - the helm chart will fail!

# Solution
As described [here](https://github.com/argoproj/argo-cd/issues/7291), the solution is to use `.Capabilities.APIVersions.Has` over `api/version/kind` instead of over just `api/version`

# Validation
Enable autoscaling on values.yml and run
```bash
helm template . --kube-version 1.21
```
Before fix the hpa would get `v2` (which does not exist on 1.21), after fix it will get `v2beta2` 

Run
```bash
helm template . --kube-version 1.24
```
Before and after fix you'll get `v2`


spencergilbert (approved):
🙇 thanks for the fix!

Issue #271, opened 2022-11-24
When running Vector in Agent mode, the Agent should tolerate all taints by default


hawkesn:
In Agent mode, it is designed to retrieve logs on all nodes via a Daemon Set. 
If there are taints on nodes, you would need to add tolerations on the Agent pod to ensure those nodes' logs are also getting grabbed.

By default, the Agent should tolerate all taints. This can be done by adding the following:
```
  tolerations:
    - operator: "Exists"
```

hawkesn:
I opened up: #272 as a PR

spencergilbert:
Hi @hawkesn, looking at other charts for applications in this space it seems as though they also default to no tolerations defined (one exception). Do you have any examples of this default?

1. [datadog](https://github.com/DataDog/helm-charts/blob/main/charts/datadog/templates/daemonset.yaml#L159-L168) - default `tolerations: []`
2. [fluentd](https://github.com/fluent/helm-charts/blob/main/charts/fluentd/templates/_pod.tpl#L90-L92) - default `tolerations: []`
3. [fluent-bit](https://github.com/fluent/helm-charts/blob/main/charts/fluent-bit/templates/_pod.tpl#L132-L135) - default `tolerations: []`
4. [promtail](https://github.com/grafana/helm-charts/blob/main/charts/promtail/templates/_pod.tpl#L99-L102) - default `tolerations: [{"effect":"NoSchedule","key":"node-role.kubernetes.io/master","operator":"Exists"},{"effect":"NoSchedule","key":"node-role.kubernetes.io/control-plane","operator":"Exists"}]`

I am trying to think of situations where you wouldn't want your telemetry collector to be on all nodes, including nodes that are being drained/cordoned - such that you can continue to ship telemetry off node for as long as possible 🤔

hawkesn:
That is an excellent question!

I dug into a few other helm charts (eg. Node Termination Handler) that run as a daemon set and, you're right, those also do not have default tolerations set.

I did find this option: `tolerateAllTaints` as a value in `aws-ebs-csi-driver`: https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/templates/node.yaml#L39-L46

This one also has no default tolerations set. Albeit, having that especially called-out in the `values.yaml` made it explicitly clear to me I should turn that on 👍.

[...]

PR #272, opened 2022-11-24
feat(vector): add default all tolerations

hawkesn:
Relates to: https://github.com/vectordotdev/helm-charts/issues/271
Adding default tolerations to the daemon set.


spencergilbert:
Hi @hawkesn I left a comment on the issue you opened for the larger discussion, specifically for this change though...

I believe this would be the default regardless of the defined `role`, if we went forward with this I think we'd want only the Agent role to have this toleration.

hawkesn:
Closing based on discussion

PR #274, opened 2022-11-30
fix(vector): Update to Vector 0.25.2

jszwedko:
Signed-off-by: Jesse Szwedko <jesse.szwedko@datadoghq.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,14 @@ All notable changes to this project will be documented in this file.
 
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html)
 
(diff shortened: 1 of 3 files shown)


neuronull (approved):
Seems to be a CI check failing though I'm not exactly sure the meaning... perhaps it is detecting a missed update to 25.2 in one of the files?

jszwedko:
> Seems to be a CI check failing though I'm not exactly sure the meaning... perhaps it is detecting a missed update to 25.2 in one of the files?

Aha, I knew I was missing something. I needed to run `helm-docs`. Thanks for catching! We probably need a wrapper script to put these all together.

PR #280, opened 2023-02-07
feat(vector): add annotations for deployment, daemonset, statefulset

enicholson125:
Would it be alright to add annotations for the pod controller of Vector? 

My specific use-case is that ArgoCD deployment doesn't work with the rollWorkload option, so I would like to be able to put the stakater/reloader annotation on the deployment to roll Vector when its ConfigMap dependencies are updated.

### DIFF
--- a/charts/vector/values.yaml
+++ b/charts/vector/values.yaml
@@ -131,6 +131,9 @@ podHostNetwork: false
 # for Vector Pods.
 podSecurityContext: {}
 
+# workloadResouceAnnotations -- Set annotations on the Vector Daemonset, Deployment or Statefulset.
+workloadResouceAnnotations: {}
+
 # securityContext -- Specify securityContext on Vector containers.
 securityContext: {}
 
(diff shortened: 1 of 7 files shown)


charts/vector/values.yaml
```diff
@@ -131,6 +131,9 @@ podHostNetwork: false
 # for Vector Pods.
 podSecurityContext: {}
 
+# workloadResouceAnnotations -- Set annotations on the Vector Daemonset, Deployment or Statefulset.
```
neuronull, line 134: Nit

```suggestion
# workloadResouceAnnotations -- Set annotations on the Vector DaemonSet, Deployment or StatefulSet.
```

neuronull (commented):
Chatted with @spencergilbert about this.

The changes look good, just one nit about the proper noun capitalization.

Would you also mind editing the test

https://github.com/vectordotdev/helm-charts/blob/develop/charts/vector/ci/aggregator-all-values.yaml

, to utilize the new annotations ?

enicholson125:
Thanks! I've added the value into the aggregators-all test and pulled in the capitalization fix.

enicholson125:
Sorry about the test failures - hopefully all working now

neuronull (approved):
LGTM!

neuronull:
Thanks for the contribution!

PR #291, opened 2023-04-01
feat(vector): add minReadySeconds for daemonsets

AndrooTheChen:
add [minReadySeconds](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/daemon-set-v1/#DaemonSetSpec:~:text=minReadySeconds%20(int32)). useful when we have a lot of nodes and we wanted to be sure the pod is fully deployed, otherwise may want to reschedule this job and investigate.


StephenWakely (approved):
Thank you. Yes this could be really useful.

PR #300, opened 2023-04-20
fix(vector): Bump Vector chart to v0.29.1

neuronull:
- fix(vector): Bump Vector chart to v0.29.1
- chore(vector): Regenerate CHANGELOG
- -chore(vector): update README

### DIFF
--- a/charts/vector/Chart.yaml
+++ b/charts/vector/Chart.yaml
@@ -18,11 +18,11 @@ maintainers:
   - name: Datadog
     email: vector@datadoghq.com
 icon: https://vector.dev/press/vector-icon.svg
-appVersion: "0.29.0-distroless-libc"
+appVersion: "0.29.1-distroless-libc"
 annotations:
   artifacthub.io/images: |
     - name: vector
-      image: timberio/vector:0.29.0-distroless-libc
+      image: timberio/vector:0.29.1-distroless-libc
     - name: haproxy
       image: haproxytech/haproxy-alpine:2.4.17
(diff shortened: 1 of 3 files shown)


charts/vector/Chart.yaml
```diff
@@ -18,11 +18,11 @@ maintainers:
   - name: Datadog
```
spencergilbert, line 18: `version` above needs bumping!

spencergilbert (changes requested):

charts/vector/Chart.yaml
neuronull: I noticed that, but chose not to do it manually because the script didn't, and in this patch PR, it was also not edited:

https://github.com/vectordotdev/helm-charts/pull/293/files

🤔

neuronull: Is that a bug in the script ?

spencergilbert: Ah, we could probably automate it. You got caught here because this is the first PR after the previous release. The linked PR didn't have to change it as it had already been bumped so it was "ready"

neuronull: Ohh. I suppose the README version needs to change as well then

spencergilbert (approved):

charts/vector/Chart.yaml
spencergilbert: Yeah, I swear that's mentioned in the docs somewhere... maybe there's release docs in confluence? Or maybe I'm thinking of OPW.

PR #302, opened 2023-05-12
Fix default/missing storageClassName

sbalmos:
Do not include storageClassName attribute if not provided (default). Currently the persistence.storageClassName value is not provided by default, which results in an empty string storageClassName attribute in the StatefulSet template. This inadvertently disables dynamic PV provisioning.

If no persistence.storageClassName value is provided, then the attribute should not be present in the rendered StatefulSet, allowing the PVC to use the cluster's default storage class.


neuronull (approved):
Thanks for catching this @sbalmos!

PR #305, opened 2023-05-25
fix(vector): move minReadySeconds to global scope

rbakhtaraev:
In this case, we must use the global (root) context for access to `.Values.minReadySeconds` variable.
Without it, Helm try to find it in `.Values.updateStrategy` scope.

closes: #306

### DIFF
--- a/charts/vector/templates/daemonset.yaml
+++ b/charts/vector/templates/daemonset.yaml
@@ -11,8 +11,8 @@ spec:
   selector:
     matchLabels:
       {{- include "vector.selectorLabels" . | nindent 6 }}
-  {{- with .Values.updateStrategy }}
   minReadySeconds: {{ .Values.minReadySeconds }}
+  {{- with .Values.updateStrategy }}
   updateStrategy:
     {{- toYaml . | nindent 4 }}
   {{- end }}
--- a/charts/vector/Chart.yaml
+++ b/charts/vector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: vector
-version: "0.22.0"
+version: "0.22.1"
 kubeVersion: ">=1.15.0-0"
 description: A lightweight, ultra-fast tool for building observability pipelines
 type: application
--- a/charts/vector/README.md
+++ b/charts/vector/README.md
@@ -1,6 +1,6 @@
 # Vector
 
-![Version: 0.22.1](https://img.shields.io/badge/Version-0.22.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 0.30.0-distroless-libc](https://img.shields.io/badge/AppVersion-0.30.0--distroless--libc-informational?style=flat-square)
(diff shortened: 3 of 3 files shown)


neuronull:
I validated this locally, thanks for dropping a fix!

neuronull:
The chart version just needs to be upgraded here: https://github.com/vectordotdev/helm-charts/blob/develop/charts/vector/Chart.yaml#L3

neuronull:
Ah, the changelog doesn't need to be modified in the individual PRs, we update it with a script as part of the release process: https://github.com/vectordotdev/helm-charts#releasing

spencergilbert (approved):

kaden-l-nelson:
@spencergilbert What're your thoughts on taking this PR instead? https://github.com/vectordotdev/helm-charts/pull/308/files

#308 makes it so that min ready seconds can be configured in the absence of an update strategy.

skygrammas:
@kaden-l-nelson makes a good point. In the absence of an update strategy value, would the [`{{- with .Values.updateStrategy }}` block](https://github.com/vectordotdev/helm-charts/blob/b2e52a3d81570faa52659c2c894255aa422b66da/charts/vector/templates/daemonset.yaml#L14) not be generated? Resulting in `minReadySeconds` not being templated despite it being reference in the root context?

spencergilbert:
> @kaden-l-nelson makes a good point. In the absence of an update strategy value, would the [`{{- with .Values.updateStrategy }}` block](https://github.com/vectordotdev/helm-charts/blob/b2e52a3d81570faa52659c2c894255aa422b66da/charts/vector/templates/daemonset.yaml#L14) not be generated? Resulting in `minReadySeconds` not being templated despite it being reference in the root context with this proposed change?

That's correct - it does look like the `with` line isn't in the right place. There doesn't need to be any connection between the `updateStrategy` and the `minReadySeconds` values so they should be templated separately.

cc @neuronull

charts/vector/templates/daemonset.yaml
```diff
@@ -12,7 +12,7 @@ spec:
     matchLabels:
       {{- include "vector.selectorLabels" . | nindent 6 }}
   {{- with .Values.updateStrategy }}
```
skygrammas, line 14:
```suggestion
   minReadySeconds: {{ $.Values.minReadySeconds }}
   {{- with .Values.updateStrategy }}
```

@rbakhtaraev would you mind swapping these lines in this proposal?

charts/vector/templates/daemonset.yaml
```diff
@@ -12,7 +12,7 @@ spec:
...
-  minReadySeconds: {{ .Values.minReadySeconds }}
+  minReadySeconds: {{ $.Values.minReadySeconds }}
```
skygrammas, line 15:
```suggestion
```

charts/vector/templates/daemonset.yaml
spencergilbert: If moved outside of the `with` block, the `$` should be dropped as well, I beleive.

kaden-l-nelson: https://github.com/vectordotdev/helm-charts/pull/308 This PR already has the correct changes.

skygrammas: This is true. @spencergilbert how do you feel about reviewing the other PR instead? Or would you prefer moving forward on this proposal?

rbakhtaraev: yep, fixed it

skygrammas: awesome, thanks.

spencergilbert:
@rbakhtaraev looks like this needs a `helm-docs` run to update the badges.

PR #307, opened 2023-05-28
Remap `host` to the node name for Agent `host_metrics`

sdwilsh:
The current behavior of the chart provides pod names of the agent to Prometheus, which isn't very helpful for host metrics.  This change is a much better default behavior, at least in my novice opinion.

### DIFF
--- a/charts/vector/templates/configmap.yaml
+++ b/charts/vector/templates/configmap.yaml
@@ -74,10 +74,15 @@ data:
         type: host_metrics
       internal_metrics:
         type: internal_metrics
+    transforms:
+      host_metrics_with_node_name:
+        type: remap
+        inputs: [host_metrics]
+        source: .tags.host = get_env_var!("VECTOR_SELF_NODE_NAME")
     sinks:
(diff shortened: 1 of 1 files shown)


spencergilbert:
This feels like a good change - but it would be a breaking change, which we haven't done with this chart yet (IIRC).

jszwedko:
We have made a couple of breaking changes: https://github.com/vectordotdev/helm-charts/blob/develop/CHANGELOG.md

Is the current behavior also potentially what some users will want? Maybe we could make it a flag?

spencergilbert:
Thinking about this - I'm not sure this is the best way to resolve. It would only apply the a default configuration, and any user-end changes would drop the use of the proper host(node)name.

Ultimately I think this should be solved in Vector itself, rather than as a step in VRL. Perhaps a short term, non-breaking change, would be documenting this VRL program as an example in the README?

spencergilbert:
I'm going to close this PR, given my thoughts above. I'd be happy to merge in a note in the documentation about this though.

PR #323, opened 2023-08-31
Ability to add extraObjects

minebaev:
Hi. This will allow users to create resources for vector such as dashboards for grafana


dsmith3197 (approved):
Small nit on the spacing but otherwise looks good.

dsmith3197:
@minebaev Could you also add `extraObjects` to https://github.com/vectordotdev/helm-charts/blob/develop/charts/vector/README.md to fix the lint error?

PR #326, opened 2023-09-06
feat: support shareProcessNamespace is pod.spec

andresperezl:
This adds the ability to set the `pod.spec.shareProcessNamespace` value in the pod template (defaults to `false`), which allows the sidecars container to be able to send the reload signal (`SIGHUP`) to the vector process.

### DIFF
--- a/charts/vector/templates/_pod.tpl
+++ b/charts/vector/templates/_pod.tpl
@@ -16,6 +16,9 @@ priorityClassName: {{ . }}
 {{- with .Values.dnsPolicy }}
 dnsPolicy: {{ . }}
 {{- end }}
+{{- with .Values.shareProcessNamespace }}
+shareProcessNamespace: {{ . }}
+{{- end }}
 {{- with .Values.dnsConfig }}
 dnsConfig:
 {{ toYaml . | indent 2 }}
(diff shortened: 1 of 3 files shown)


charts/vector/templates/_pod.tpl
```diff
@@ -16,6 +16,9 @@ priorityClassName: {{ . }}
 {{- with .Values.dnsPolicy }}
 dnsPolicy: {{ . }}
 {{- end }}
+{{- with .Values.shareProcessNamespace }}
+shareProcessNamespace: {{ . }}
+{{- end }}
 {{- with .Values.dnsConfig }}
 dnsConfig:
 {{ toYaml . | indent 2 }}
```
dsmith3197, line 24: Small nit - can we keep `dnsPolicy` and `dnsConfig` next to each other as they are related?

```suggestion
{{- with .Values.shareProcessNamespace }}
shareProcessNamespace: {{ . }}
{{- end }}
{{- with .Values.dnsPolicy }}
dnsPolicy: {{ . }}
{{- end }}
{{- with .Values.dnsConfig }}
dnsConfig:
{{ toYaml . | indent 2 }}
```

dsmith3197 (changes requested):
Can you also update `charts/vector/README.md` with the new `shareProcessNamespace` option?

dsmith3197 (approved):

PR #337, opened 2023-10-17
Remove defaults for PDB

bhataprameya:
- Remove default for PDB. This allows the user to to set only the values required. 
- Currently they add both both `minAvailable` and `maxUnavailable` are added even of I just want to set `maxUnavailable`.
- lot of k8s deployments only allow one of this 2 values to be passed causing the deployment to fail

### DIFF
--- a/charts/vector/values.yaml
+++ b/charts/vector/values.yaml
@@ -88,9 +88,9 @@ podDisruptionBudget:
   # for Vector.
   enabled: false
   # podDisruptionBudget.minAvailable -- The number of Pods that must still be available after an eviction.
-  minAvailable: 1
+  # minAvailable: 1
   # podDisruptionBudget.maxUnavailable -- (int) The number of Pods that can be unavailable after an eviction.
-  maxUnavailable:
+  # maxUnavailable:
 
 rbac:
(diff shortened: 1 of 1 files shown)


charts/vector/values.yaml
```diff
@@ -88,9 +88,9 @@ podDisruptionBudget:
   # for Vector.
   enabled: false
   # podDisruptionBudget.minAvailable -- The number of Pods that must still be available after an eviction.
-  minAvailable: 1
+  # minAvailable: 1
   # podDisruptionBudget.maxUnavailable -- (int) The number of Pods that can be unavailable after an eviction.
-  maxUnavailable:
+  # maxUnavailable:
```
dsmith3197, line 93: Unfortunately this is a breaking change to the Helm chart, which will impact other users.

Are you able to get your setup to work with the following configuration?
```
  enabled: true
  # podDisruptionBudget.minAvailable -- The number of Pods that must still be available after an eviction.
  minAvailable:
  # podDisruptionBudget.maxUnavailable -- (int) The number of Pods that can be unavailable after an eviction.
  maxUnavailable: 1
```

dsmith3197:
I'm going to close this for now given the suggestion above.  Please re-open this PR if this needs further attention.

PR #346, opened 2023-11-17
fix: rendering custom config tpl

kimxogus:
There's `{{ tpl (toYaml .Values.customConfig) . | indent 4 }}`, but we can't actually use tpl becuase there's several points accessing and iterating child objects of `.Values.customConfig`.
So I made them not to access or iterate child objects if customConfig is string.

### DIFF
--- a/charts/vector/README.md
+++ b/charts/vector/README.md
@@ -1,6 +1,6 @@
 # Vector
 
-![Version: 0.34.0](https://img.shields.io/badge/Version-0.34.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 0.39.0-distroless-libc](https://img.shields.io/badge/AppVersion-0.39.0--distroless--libc-informational?style=flat-square)
+![Version: 0.34.1](https://img.shields.io/badge/Version-0.34.1-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 0.39.0-distroless-libc](https://img.shields.io/badge/AppVersion-0.39.0--distroless--libc-informational?style=flat-square)
 
 [Vector](https://vector.dev/) is a high-performance, end-to-end observability data pipeline that puts you in control of your observability data. Collect, transform, and route all your logs, metrics, and traces to any vendors you want today and any other vendors you may want tomorrow. Vector enables dramatic cost reduction, novel data enrichment, and data security where you need it, not where is most convenient for your vendors.
 
--- a/charts/vector/templates/_helpers.tpl
+++ b/charts/vector/templates/_helpers.tpl
@@ -176,10 +176,6 @@ Generate an array of ContainerPorts based on `.Values.customConfig`.
       {{- end }}
     {{- end }}
   {{- end }}
-{{- else if or .Values.livenessProbe .Values.readinessProbe }}
-- name: api
-  containerPort: 8686
-  protocol: TCP
 {{- end }}
 {{- end }}
 
--- a/charts/vector/templates/_helpers.tpl
+++ b/charts/vector/templates/_helpers.tpl
@@ -162,6 +162,7 @@ Generate a single ServicePort based on a component configuration.
 Generate an array of ContainerPorts based on `.Values.customConfig`.
 */}}
 {{- define "vector.containerPorts" -}}
+{{- if not (typeIs "string" .Values.customConfig) }}
   {{- range $componentKind, $components := .Values.customConfig }}
     {{- if eq $componentKind "sources" }}
       {{- tuple $components "_helper.generateContainerPort" | include "_helper.componentIter" }}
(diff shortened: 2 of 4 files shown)


kimxogus:
Can anyone review this?

kimxogus:
@dsmith3197 Is there any progress reviewing this pr?

dsmith3197 (commented):
I'm concerned that the current approach breaks the automatic detection of container ports and the data directory when using templates in the Vector config. Is there a way we can workaround this?

kimxogus:
@dsmith3197 Yes, this could break those container port related templates, but I think adding all those additional values for container ports will make this pr too large

kimxogus:
@dsmith3197 Is this ok?

jszwedko:
Hi @kimxogus ,

Apologies in the delay of review of this. I'm admittedly having trouble with the whitespace trying to test this out but templating like this should be fine for templating the list value rather than a string, no? I'm looking at https://helm.sh/docs/chart_template_guide/control_structures/ which shows an example of doing it to add a key/value to a map.

kimxogus:
@jszwedko I just wanted to fix the [broken tpl config](https://github.com/vectordotdev/helm-charts/blob/develop/charts/vector/templates/configmap.yaml#L11).
Also, templating in string config let you handle different configs in single template. as I mentioned [above](https://github.com/vectordotdev/helm-charts/pull/346#issuecomment-1837847346).
Using list or key-value config, you have to copy and paste your values.yaml everywhere or write a very complicated templates and conditions in your helm chart.

kimxogus:
@jszwedko Sorry, I was confused the context as it's 4month old.
The original intention was to support string type `customConfig` because it's the best way to utilize tpl.
In current helm templates, is there any way to implement conditional features like below?
We are using our own helm chart wrapping vector's helm chart with those type of helm template, and deployed in various environments.
I believe this kind of pattern is the most reusable and convenient way to manage various environments with a single template.
If not, we have to write different values.yaml and edit in every helm releases(like adding every single topic names and additional configs in each values.yaml), which is not reusable and make us too hard to manage our data pipelines with vector.
```yaml
global:
  feature:
    a: true
    b: false

vector:
  customConfig: |
    sources:
      kafka:
        type: kafka
        bootstrap_servers: KAFKA:9092
        decoding:
          codec: json
        group_id: MY_GROUP
        topics:
          - common-topic
          {{- if .Values.global.feature.a }}
          - feature-a-topic
          {{- end }}
          {{- if .Values.global.feature.b }}
          - feature-b-topic
          {{- end }}
```

jszwedko:
Thanks @kimxogus ! My understanding is that it should be possible to template the YAML as-is like I mentioned in https://github.com/vectordotdev/helm-charts/pull/346#issuecomment-2071065355 (though admittedly I was still struggling with the formatting). Would that work for you?

jszwedko:
E.g. like they show here:

```
data:
  myvalue: "Hello World"
  drink: {{ .Values.favorite.drink | default "tea" | quote }}
  food: {{ .Values.favorite.food | upper | quote }}
  {{ if eq .Values.favorite.drink "coffee" }}mug: "true"{{ end }}
```

kimxogus:
The code `{{ tpl (toYaml .Values.customConfig) . | indent 4 }}` converts the yaml to string first, and then runs the `tpl` function. That means `{{ if eq .Values.favorite.drink "coffee" }}mug: "true"{{ end }}` won't work in `toYaml` function, before `tpl` because it's go template format, not valid yaml format.

Helm raised this error with the template.
`Error: cannot load values.yaml: error converting YAML to JSON: yaml: line ___: did not find expected key`

I think

jszwedko:
@kimxogus Ah I see. Would it also work to run the `tpl` function on the string before doing `toYaml` then?

kimxogus:
@jszwedko No, as `.Values.customConfig` is not string

kimxogus:
Yes

jszwedko:
> Yes

👍 do you mind updating this PR to do that instead, then?

kimxogus:
@jszwedko  I've just helm read documents. `tpl` function generates string output, and `toYaml` function converts key-value map to string. `tpl` can't be run before `toYaml`.
`toYaml` should be before `tpl` as [here](https://github.com/vectordotdev/helm-charts/pull/346/files#diff-0ff558f0e281ff800cac50deebc02587a1d719fac5809b712a074976abec68afR14).

DingGGu:
Any updates? @jszwedko

kimxogus:
@jszwedko Hi, `toYaml`'s input is object and `tpl` consume and produces string, so it's not possible to run `tpl` before `toYaml`. Can we merge this?

jszwedko:
Hey! Apologies, i still have to circle back to this PR. I'm still not really sure I understand the why change is necessary here but that is likely due to my lack of familiarity with Helm. I'll try to spend more time with this soon.

pront:
Looks abandoned. Feel free to reopen for any reason.

PR #347, opened 2023-11-24
enhancement(vector): Add `persistence.retentionPolicy` for configuring Statefulset PVC retention policy

alemuro:
Hello! 👋 

This PR adds a new variable in the `values.yaml` file: `persistence.retentionPolicy`. 

This allow users to define which is the PVC retention policy for the created Statefulsets. Kubernetes by default sets a `Remain` policy, meaning that no PVC are removed when the Statefulset is down-scaled or removed.  With this new setting, users can tell Kubernetes to remove these disks, for example for non production environment. 

This variable has a default set to `{}`, meaning that this setting will not be applied and Kubernetes will apply its default value, so current statefulsets will not be affected.

More info about this new feature here:
https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#persistentvolumeclaim-retention

## Some tests

I've run the following commands to ensure that the output is expected:

**Using this setting on a K8S cluster below 1.23**: Setting do not appear on the rendered statefulset.

```shell
$ helm template \                        
      -s templates/statefulset.yaml  \
      --kube-version "1.22" --set 'persistence.retentionPolicy.whenDeleted=Delete' \
      . |                  
      yq -r '.spec.persistentVolumeClaimRetentionPolicy'
null
```

**Using this setting on a k8s cluster above 1.23**: Setting appears on the rendered statefulset.

```shell
$ helm template \
      -s templates/statefulset.yaml  \
      --kube-version "1.27" --set 'persistence.retentionPolicy.whenDeleted=Delete' \
      . |
      yq -r '.spec.persistentVolumeClaimRetentionPolicy'
{
  "whenDeleted": "Delete"
}

$ helm template \
      -s templates/statefulset.yaml  \
      --kube-version "1.27" \
      --set 'persistence.retentionPolicy.whenScaled=Delete' \
      . |
      yq -r '.spec.persistentVolumeClaimRetentionPolicy'
{
  "whenScaled": "Delete"
}
```

**Using the default value in a k8s cluster above 1.23**: Setting do not appear on the rendered statefulset.

```shell
$ helm template \
      -s templates/statefulset.yaml  \
      --kube-version "1.27" \                                                                            
      . |
      yq -r '.spec.persistentVolumeClaimRetentionPolicy'
null
```

### DIFF
--- a/charts/vector/values.yaml
+++ b/charts/vector/values.yaml
@@ -365,20 +365,11 @@ persistence:
   # "Aggregator" role.
   # storageClassName: default
 
-  # The [Persistent Volume Claim (PVC) retention policy](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#persistentvolumeclaim-retention)
-  # controls if and how PVCs are deleted during the lifecycle of a StatefulSet.
-  # WhenDeleted specifies what happens to PVCs created from StatefulSet VolumeClaimTemplates when the StatefulSet is deleted,
-  # and WhenScaled specifies what happens to PVCs created from StatefulSet VolumeClaimTemplates when the StatefulSet is scaled down.
-  #
-  # Example:
-  #
-  # ```yaml
-  # retentionPolicy:
-  #   whenDeleted: Retain
-  #   whenScaled: Retain
-  # ```
-  # @type: map
+  # persistence.retentionPolicy -- Configure a [PersistentVolumeClaimRetentionPolicy](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#persistentvolumeclaim-retention)
+  # for Vector's PersistentVolumeClaims. Valid for the "Aggregator" role.
   retentionPolicy: {}
+  #  whenDeleted: Retain
+  #  whenScaled: Retain
 
   # persistence.accessModes -- Specifies the accessModes for PersistentVolumeClaims. Valid for the "Aggregator" role.
   accessModes:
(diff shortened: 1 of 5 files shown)


charts/vector/values.yaml
```diff
@@ -365,6 +365,21 @@ persistence:
...
+  # 
+  # Example:
+  #
+  # ```yaml
+  # persistentVolumeClaimRetentionPolicy:
+  #   whenDeleted: Retain
+  #   whenScaled: Retain
+  # ```
+  # @type: map
+  persistentVolumeClaimRetentionPolicy: {}
```
dsmith3197, line 381: `persistence.persistentVolumeClaimRetentionPolicy` is quite lengthy and redundant. What do you think about naming this `persistence.retentionPolicy` instead?

dsmith3197 (changes requested):
This looks good overall. Thank you for testing the changes!

Could you please add a CHANGELOG entry, bump the chart version, and change the PR title to follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)? ([ref](https://github.com/vectordotdev/helm-charts/pull/336#issuecomment-1815024218))

charts/vector/values.yaml
alemuro: Sure, will do it! The reason of using the whole name is because this is the name of the variable in Kubernetes, but it might be redundant as you said 👍

alemuro: Hello again! I've executed the requested actions, but according to the README, I think they should be executed on `develop` after merging into `main`. 

- Chart is using the latest Vector version, so I didn't bumped it.
- Updated the Chart version in Chart.yaml from `0.29.0` to `0.30.0`.
- Updated Changelog manually: I've created a new commit (otherwise I can't reference the commit itself in the Changelog)
- I've renamed the variable name as proposed
- I've updated the tests in the description (and I've run them again to validate that it works fine)
- I've updated the commit to include the PR and to be a conventional commit ( `enhance(): ...`)

Hope it's fine now! Please let me know if I should run any other action 🙏 

Many thanks 😄

dsmith3197 (approved):

jszwedko (approved):

PR #367, opened 2024-02-09
fix(vector): add minReadySeconds to StatefulSet and Deployment specs

alexanderldavis:
The chart currently supports `minReadySeconds` for DaemonSet-based deployments of the Vector agent. This PR extends this value to also apply to StatefulSets and Deployment specs.

### DIFF
--- a/charts/vector/README.md
+++ b/charts/vector/README.md
@@ -164,7 +164,7 @@ helm install --name <RELEASE_NAME> \
 | lifecycle | object | `{}` | Set lifecycle hooks for Vector containers. |
 | livenessProbe | object | `{}` | Override default liveness probe settings. If customConfig is used, requires customConfig.api.enabled to be set to true. |
 | logLevel | string | `"info"` |  |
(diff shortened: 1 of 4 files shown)


jszwedko (approved):
Thanks @alexanderldavis . This change makes sense to me!

jszwedko:
@alexanderldavis it looks like there are some CI failures here when you get some time.

alexanderldavis:
@jszwedko I'm struggling to find a good way to pass the linter for older kube versions. `minReadySeconds` was added to Statefulsets in v1.22, but made generally available in v1.25. Do you have a recommendation for how that should be handled?

jszwedko:
@alexanderldavis I was expecting the failing CI checks to block the automerge but apparently they didn't 😓 I opened https://github.com/vectordotdev/helm-charts/pull/371 to revert.

> minReadySeconds was added to Statefulsets in v1.22, but made generally available in v1.25

My assumption would be that we'd only want to template it out for v1.25+ then, but maybe I'm missing something 🤔

PR #384, opened 2024-04-09
fix(vector): Bump Vector to v0.37.1

jszwedko:


charts/vector/Chart.yaml
```diff
@@ -18,11 +18,11 @@ maintainers:
   - name: Datadog
     email: vector@datadoghq.com
 icon: https://vector.dev/press/vector-icon.svg
-appVersion: "0.37.0-distroless-libc"
+appVersion: "0.37.1-distroless-libc"
 annotations:
   artifacthub.io/images: |
     - name: vector
-      image: timberio/vector:0.37.0-distroless-libc
+      image: timberio/vector:0.37.1-distroless-libc
```
tobz, line 25: Should we, at some point, actually move this to a dedicated `vectordotdev` or `vector` org so avoid the Timber.io baggage?

tobz (approved):

charts/vector/Chart.yaml
jszwedko: Yeah, we do plan to move this over to `vectordotdev` at some point 🙏

PR #385, opened 2024-04-30
fix(vector): have psp support hostNetwork: true when podHostNetwork is true

nascd:


jszwedko (approved):
Thanks @nascd !

jszwedko:
I think you just need to bump the chart version.

jszwedko:
(pressed the wrong button 😓 )

nascd:
> I think you just need to bump the chart version.

Thanks @jszwedko, just a patch increment to https://github.com/vectordotdev/helm-charts/blob/develop/charts/vector/Chart.yaml#L3 ?  Should I include that in this PR, or a follow-up after this is merged?

jszwedko:
> > I think you just need to bump the chart version.
> 
> Thanks @jszwedko, just a patch increment to https://github.com/vectordotdev/helm-charts/blob/develop/charts/vector/Chart.yaml#L3 ? Should I include that in this PR, or a follow-up after this is merged?

Yeah, I think it would just be a patch bump. It'll have to be part of this PR to satisfy the linter 🙏

nascd:
@jszwedko I've bumped the version.  I don't appear to have permissions to merge, but please let me know if there's anything else I need to do.  Thank you!

jszwedko (approved):
Thanks!

nascd:
Is there something else required to get this chart available in the public repository?

jszwedko:
> Is there something else required to get this chart available in the public repository?

We are planning a release for next week so I think we can just loop it in with that.

PR #386, opened 2024-05-02
enhancement(vector): Allow external autoscaler

srstrickland:
When using any autoscaler, replicas should not be set on any Deployment or StatefulSet. This helm chart has hooks for defining an HPA, but there are times when users may want to create their own, or use something external like [KEDA](https://keda.sh). In our case, we want to use KEDA for scaling based on kafka lag and other triggers. This largely works as expected, but I need to prevent the `replicas` field from being populated, or else things appear out-of-sync when the external autoscaler is doing its thing (we use ArgoCD and it will show the project as out-of-sync because of replicas).

The feature here is just to offer an additional field under `autoscaling` options: `external: true/false`.

### DIFF
--- a/charts/vector/Chart.yaml
+++ b/charts/vector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: vector
-version: "0.32.2"
+version: "0.33.0"
 kubeVersion: ">=1.15.0-0"
 description: A lightweight, ultra-fast tool for building observability pipelines
 type: application
--- a/charts/vector/README.md
+++ b/charts/vector/README.md
@@ -1,6 +1,6 @@
 # Vector
 
-![Version: 0.32.1](https://img.shields.io/badge/Version-0.32.1-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 0.37.1-distroless-libc](https://img.shields.io/badge/AppVersion-0.37.1--distroless--libc-informational?style=flat-square)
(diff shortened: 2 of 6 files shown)


jszwedko (approved):
Thanks for this contribution @srstrickland ! I think you just need to bump the chart version to satisfy the linter in CI.

srstrickland:
Ah, great... will do.  Should I put in a note in the `CHANGELOG.md` as well?  Side note: I see the current chart version is 0.32.2, but only 0.32.1 has been produced (and no note in changelog about 0.32.2).  So I wasn't sure if you had an external release cycle.  Let me know, happy to update anything.

jszwedko:
> Ah, great... will do. Should I put in a note in the `CHANGELOG.md` as well? Side note: I see the current chart version is 0.32.2, but only 0.32.1 has been produced (and no note in changelog about 0.32.2). So I wasn't sure if you had an external release cycle. Let me know, happy to update anything.

Ah, I missed the version was updated in another PR. This is a feature addition rather than a bug fix, though, so we should update to 0.33.0 to follow semvar. The changelog is generated during release, from commit titles, so you don't have to update that.

jszwedko (approved):

srstrickland:
Any thoughts on when 0.33.0 will be published?  It's not blocking me; I've published this update to our own internal repo, just would prefer to use the official one whenever it's available.

jszwedko:
> Any thoughts on when 0.33.0 will be published? It's not blocking me; I've published this update to our own internal repo, just would prefer to use the official one whenever it's available.

Should be this week!

PR #388, opened 2024-05-09
enhancement(vector): Update the image fields to allow for a static base name

### LINKED ISSUE #237: Reintroduce `base` for image config
Hi 👋🏼 

I saw this was removed with the switch to the new vector chart. I am running vector agent with the debian flavor of the image because it includes `journalctl`

```yaml
image:
  repository: timberio/vector
  tag: 0.23.3-debian
```

It would be nice if we could set a `flavor` or `base` config defining the image type we want to use:

```yaml
image:
  repository: timberio/vector
  base: debian # or flavor: debian
```

The reason I see this being helpful is that as a consumer of the chart when I go to update the chart version, I also need to update the `values.yaml` file with the new vector image. It would be nice to only care about the chart version instead of the chart version and image tag.

Thanks!

jandersen-plaid:
This fixes #237 by allowing the user of the chart to set a static `.Values.image.base` field.

When this is set, it will be used to replace the `base` that is used in the `AppVersion` where appropriate, so that users can upgrade their chart without having to constantly change the image tag to keep the version up.

The order of precedence for this is:
1. `.Values.image.tag` overrides everything
2. `.Values.image.base` overrides the base of the `.Chart.AppVersion`
3. `.Chart.AppVersion` is the default

I could not for the life of me get chart testing to work locally for some reason, but I am happy to update the README.md by hand if necessary.

### DIFF
--- a/charts/vector/Chart.yaml
+++ b/charts/vector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: vector
-version: "0.33.1"
+version: "0.34.0"
 kubeVersion: ">=1.15.0-0"
 description: A lightweight, ultra-fast tool for building observability pipelines
 type: application
--- a/charts/vector/Chart.yaml
+++ b/charts/vector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: vector
-version: "0.33.0"
+version: "0.33.1"
 kubeVersion: ">=1.15.0-0"
 description: A lightweight, ultra-fast tool for building observability pipelines
 type: application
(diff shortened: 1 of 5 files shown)


jszwedko:
I'm curious what you saw when you ran `helm-docs` but I pushed a commit regenerating the README.md to satisfy CI.

jandersen-plaid:
> I'm curious what you saw when you ran `helm-docs` but I pushed a commit regenerating the README.md to satisfy CI.

I think my local setup is just a bit borked -- I will need to fix it later today, but I did not want to slow down on getting the PR up while I had the motivation and time. Thank you for the review!

PR #394, opened 2024-06-21
fix(vector): Add `namespace` to manifests

mustdiechik:
To prevent argo CD InvalidSpecError

### DIFF
--- a/charts/vector/Chart.yaml
+++ b/charts/vector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: vector
-version: "0.34.0"
+version: "0.34.1"
 kubeVersion: ">=1.15.0-0"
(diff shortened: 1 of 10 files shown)


jszwedko (commented):
Thanks @mustdiechik ! I think you'll need to bump the chart version.

jszwedko (approved):
Thanks @mustdiechik

jszwedko:
Just an FYI that I had to revert this since it caused a CI failure:

```
>>> helm lint charts/vector --values charts/vector/ci/datadog-values.yaml
==> Linting charts/vector
Error:  templates/: template: vector/templates/secret.yaml:6:24: executing "vector/templates/secret.yaml" at <.Release.Namespace>: nil pointer evaluating interface {}.Namespace
```

PR #402, opened 2024-08-07
docs(README): intended recent available links

guspan-tanadi:


jszwedko (approved):
Thanks @guspan-tanadi , I'm not sure why it was previously referring to latest.

Issue #391, opened 2024-05-18
add value hostAliases


VitalyEst:
Please add value hostAliases to 
https://github.com/vectordotdev/helm-charts/blob/develop/charts/vector/templates/_pod.tpl

```
{{- with .Values.hostAliases }}
hostAliases:
{{ toYaml . | indent 2 }}
{{- end }}
```

https://github.com/vectordotdev/helm-charts/blob/develop/charts/vector/values.yaml

```
hostAliases: []
#- ip: "127.0.0.1"
#  hostnames:
#  - "foo.local"
#  - "bar.local"
```

VitalyEst:
This parameter used for connect network resources without DNS service.

jszwedko:
👍 we'd support a PR here if anyone is motivated.

PR #403, opened 2024-08-19
fix(vector): add hostAliases in pod spec

pigletfly:
fix of https://github.com/vectordotdev/helm-charts/issues/391


jszwedko (approved):
Thanks @pigletfly !

PR #405, opened 2024-08-27
fix(vector): internalTrafficPolicy on service

### LINKED ISSUE #399: internalTrafficPolicy does not apply for headless services
https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/2086-service-internal-traffic-policy/README.md#proposal

> The internalTrafficPolicy field will not apply for headless Services or Services of type ExternalName.

It would seem the change made in #320 is a no-op. It would function correctly if added [here](https://github.com/vectordotdev/helm-charts/blob/72ffb20576cf4141b6881ada2660fdaadf2e22ad/charts/vector/templates/service.yaml#L13) for the ClusterIP service.

H4nks:
Fixes #399

### DIFF
--- a/charts/vector/templates/service-headless.yaml
+++ b/charts/vector/templates/service-headless.yaml
@@ -18,9 +18,6 @@ spec:
 {{- if .Values.service.ipFamilyPolicy }}
   ipFamilyPolicy: {{ .Values.service.ipFamilyPolicy }}
 {{- end }}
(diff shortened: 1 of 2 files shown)


jszwedko (approved):
Good catch, thanks @H4nks !

H4nks:
I've updated the chart version @jszwedko. Should be good to go now 🙂

Issue #418, opened 2024-10-17
Publish helm charts to OCI (ex: ghcr)


domolitom:
For having the benefit of Flux automatic artifact update it would be beneficial to publish the chart also to an OCI registry.

PR #419, opened 2024-10-17
feat: add OCI registry support for Helm chart

domolitom:
Issue: https://github.com/vectordotdev/helm-charts/issues/418

Before merging this PR, probably an admin needs to permit this repository on Github Packages to have permission to push.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -6,7 +6,7 @@ on:
     branches:
       - master
     paths:
-      - 'charts/**'
+      - "charts/**"
 
 jobs:
   release:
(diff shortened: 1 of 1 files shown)


jszwedko (commented):
Thanks @domolitom ! I think we need to add permissions to push to GHR via:

```yaml
permissions:
  packages: write
```

Per https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token

domolitom:
@jszwedko good point! Sorry about that :) It is fixed now.
+ As I know an admin needs to permit this registry to write to packages.

jszwedko (approved):
Thanks @domolitom ! We are releasing Vector v0.42.0 today so we can see if this works as expected.

jszwedko:
@domolitom the latest release was published to ghcr! Thanks again for this.

PR #427, opened 2024-11-26
Update configmap.yaml

k-harutyunyan:
This condition checks for the customConfig field for the vector config not the haproxy, hence when you have a customConfig set for vector and enable haproxy you get empty haproxy config.


pront (approved):
Tanks @k-harutyunyan

PR #428, opened 2024-11-26
chore(dev): add .gitignore

pront:


charts/vector/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: vector
-version: "0.37.0"
+version: "0.38.0"
```
pront, line 3: This also unblocks https://github.com/vectordotdev/helm-charts/pull/427

bfung (approved):

PR #435, opened 2024-12-10
fix(vector): Allow empty workloadResourceAnnotations

nebula-it:
This prevents from settings empty annotations like `annotations: {}` which causes argocd to always see DaemonSet as out of sync, because a DaemonSet will get `deprecated.daemonset.template.generation` annotation anyways


jszwedko (approved):
Thanks @nebula-it !

PR #449, opened 2025-02-06
fix(vector): Use `.KubeVersion.Version` rather than `.KubeVersion.GitVersion` for capability detection

jagrusy:
I've been running into issues deploying newer versions of Vector a custom release tool due to the use of `GitVersion` which is an undocumented Helm behavior described in https://github.com/helm/helm/issues/12072. This changes a few conditionals to just user `Version`

### DIFF
--- a/charts/vector/Chart.yaml
+++ b/charts/vector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: vector
-version: "0.40.0"
+version: "0.40.1"
 kubeVersion: ">=1.15.0-0"
 description: A lightweight, ultra-fast tool for building observability pipelines
 type: application
--- a/charts/vector/templates/ingress.yaml
+++ b/charts/vector/templates/ingress.yaml
@@ -1,12 +1,12 @@
 {{- if .Values.ingress.enabled -}}
-{{- if and .Values.ingress.className (not (semverCompare ">=1.18-0" .Capabilities.KubeVersion.GitVersion)) }}
+{{- if and .Values.ingress.className (not (semverCompare ">=1.18-0" .Capabilities.KubeVersion.Version)) }}
   {{- if not (hasKey .Values.ingress.annotations "kubernetes.io/ingress.class") }}
   {{- $_ := set .Values.ingress.annotations "kubernetes.io/ingress.class" .Values.ingress.className}}
   {{- end }}
 {{- end }}
-{{- if semverCompare ">=1.19-0" .Capabilities.KubeVersion.GitVersion -}}
+{{- if semverCompare ">=1.19-0" .Capabilities.KubeVersion.Version -}}
 apiVersion: networking.k8s.io/v1
-{{- else if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
+{{- else if semverCompare ">=1.14-0" .Capabilities.KubeVersion.Version -}}
 apiVersion: networking.k8s.io/v1beta1
 {{- else -}}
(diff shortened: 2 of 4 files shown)


charts/vector/templates/ingress.yaml
```diff
@@ -1,12 +1,12 @@
 {{- if .Values.ingress.enabled -}}
-{{- if and .Values.ingress.className (not (semverCompare ">=1.18-0" .Capabilities.KubeVersion.GitVersion)) }}
+{{- if and .Values.ingress.className (not (semverCompare ">=1.18-0" .Capabilities.KubeVersion.Version)) }}
```
pront, line 2: Thank you for this fix @jagrusy. Is `GitVersion` equivalent to `Version`? I wonder if this is a backwards compatible change.

taliastocks: `GitVersion` isn't equivalent to anything, it doesn't exist in the documentation https://helm.sh/docs/chart_template_guide/builtin_objects/

pront (approved):

pront:
Needs a version bump: https://github.com/vectordotdev/helm-charts/actions/runs/13185066377/job/36807919009?pr=449

charts/vector/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: vector
-version: "0.40.0"
+version: "0.40.1"
```
pront, line 3: We want `0.41.0` here, right @jszwedko?

jszwedko: It depends on if this is a bug fix or a new feature. I can't quite tell looking at it, but it seems like a fix that doesn't change behavior. Can you confirm @jagrusy ?

The PR title also needs to be adjusted to match "semantic commit" conventions too, otherwise it won't appear in the change log. For example: `fix(vector): Use `.KubeVersion.Version` rather than `.KubeVersion.GitVersion` for capability detection`.

jagrusy: Correct there is not behavior change. I went ahead and updated the PR title. Thanks!

pront: Thank you both. Let's keep `0.40.1` then. One more tweak needed: https://github.com/vectordotdev/helm-charts/actions/runs/13187461471/job/36814871467?pr=449

jagrusy: @jszwedko is there a script to increment the version in the proper places? looks like the lint failed since there's another spot in the readme that needs updated. want to make sure I don't miss anything else.

pront: I believe you just need to run `helm-docs`.

jszwedko: Ah, yeah, it does look like `helm-docs` is needed. We could make this smoother here (scripting or at least documentation).

jagrusy: awesome, just ran it

pront: 👍 https://github.com/vectordotdev/helm-charts/issues/450

jagrusy: Hey y'all, I was away for a bit, but wanted to circle back here. Could I get an approval for the workflows to run?

jagrusy:
Hi @pront wanted to circle back here. Are you able to approve the workflow run?

pront:
Thank you @jagrusy!

jagrusy:
@pront thanks for the help and the merge! any idea when a release will be created?

PR #460, opened 2025-02-24
chore(releasing): Update Vector version to 0.45.0

pront:
Ref: https://github.com/vectordotdev/vector/issues/22445


jszwedko (commented):
I think this needs to bump the chart version too. And the title isn't quite right :) It is updating Vector version to 0.45.0.

jszwedko (approved):

PR #464, opened 2025-03-03
fix(vector): Quote namespace strings

robbie-demuth:
This prevents errors when namespaces are "numeric"

### DIFF
--- a/charts/vector/Chart.yaml
+++ b/charts/vector/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: vector
-version: "0.41.0"
+version: "0.41.1"
 kubeVersion: ">=1.15.0-0"
 description: A lightweight, ultra-fast tool for building observability pipelines
 type: application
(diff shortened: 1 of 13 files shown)


pront (approved):
Thanks @robbie-demuth

pront:
Please bump the version to fix the following:
```
 ✖︎ vector => (version: "0.41.0", path: "charts/vector") > chart version not ok. Needs a version bump! 
```

robbie-demuth:
Done, @pront. Thanks for the speedy review!

pront:
@robbie-demuth, one question. Will this always introduce extra quotes for string values? E.g. from `foo` -> `"foo"`?

robbie-demuth:
> @robbie-demuth, one question. Will this always introduce extra quotes for string values? E.g. from `foo` -> `"foo"`?

Yes, but that won't cause any harm. In fact, it's recommended as a best practice

> The easiest way to avoid type conversion errors is to be explicit about strings, and implicit about everything else. Or, in short, _quote all strings_.

https://helm.sh/docs/chart_best_practices/values/#make-types-clear

PR #469, opened 2025-03-27
feat(vector):  Support tpl for init and extra containers.

JaredTan95:
support tpl in extraContainers and initContainers.

### DIFF
--- a/charts/vector/templates/_pod.tpl
+++ b/charts/vector/templates/_pod.tpl
@@ -32,7 +32,8 @@ hostAliases:
 {{ toYaml . | indent 2 }}
 {{- end }}
 {{- if .Values.initContainers }}
-  {{- tpl (toYaml .Values.initContainers) . | nindent 2 }}
+initContainers:
+{{- tpl (toYaml .Values.initContainers) . | nindent 2 }}
 {{- end }}
 containers:
   - name: vector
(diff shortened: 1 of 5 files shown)


charts/vector/templates/_pod.tpl
```diff
@@ -31,9 +31,9 @@ imagePullSecrets:
 hostAliases:
 {{ toYaml . | indent 2 }}
 {{- end }}
-{{- with .Values.initContainers }}
+{{- if .Values.initContainers }}
 initContainers:
-{{ toYaml . | indent 2 }}
+{{- tpl (toYaml .Values.initContainers) . | nindent 2 }}
```
pront, line 36: I am not familiar wit this area and I wonder if this is a breaking change? Is it valid for users to rely on `.Values.initContainers` being treated as plain text?

JaredTan95: It's not breaking changes, it's compatible with what was there before, plain text can also be parsed normally

pront (approved):
Thanks @JaredTan95

JaredTan95:
hi @pront ,May I ask how long is the release cycle? When will this PR content be released?

pront:
Targeting Monday 7th
https://calendar.google.com/calendar/event?action=TEMPLATE&tmeid=YjQ3ajE4NnM5YjNybGxzY2xoaHFzcDhkaWtfMjAyNTA0MDcgY19xc3ExbnJzZ2V0c3Nwa243cGpycWg2Y2NsY0Bn&tmsrc=c_qsq1nrsgetsspkn7pjrqh6cclc%40group.calendar.google.com&scp=ALL

PR #471, opened 2025-04-08
feat(vector): Update Vector version to 0.46.0

pront:
Ref: https://github.com/vectordotdev/vector/issues/22784


jszwedko (approved):
I updated the PR title since we want this change to appear in the changelog (`chore`s don't).

PR #511, opened 2025-10-24
feat(vector): allow templating pod annotations

elohmeier:
Workload templates now render podAnnotations through tpl, so users can supply templated checksums (e.g. hashing external ConfigMaps) without chart changes. This is useful when no cluster access is available e.g. when using the rendered manifest pattern.


pront (approved):
Looks reasonable. 

Can this theoretically break someone if they had literal double-brace strings in their annotations?

```
  podAnnotations:
    example: "This is a {{ literal string }}"  # Would now be interpreted as template
```

pront:
Hi @elohmeier, apologies for the long delay on this one. We had a sudden surge of more urgent work. Did you close this PR because it's no longer needed? If still needed, happy to discuss more.

PR #536, opened 2026-03-02
fix(vector): use dedicated annotations for headless service

iarlyy:
Context:
The headless Service template uses `service.annotations`. This means any annotations set on the main Service leak into the headless Service, which causes conflicts with controllers such as the AWS LBC trying to reconcile a Service with load balancer config.

Proposed changes:

- serviceHeadless gets its own annotations


pront (commented):
Hi @iarlyy, thank you for the PR. Is this a breaking change for users who currently rely on `service.annotations` applying to both services?

pront:
> > Hi @iarlyy, thank you for the PR. Is this a breaking change for users who currently rely on `service.annotations` applying to both services?
> 
> Hi @pront,
> 
> yes, this is a breaking change from the point of view that existing annotations will be removed if the user doesn't define them under the `serviceHeadless` config, though I don't expect anything around vector itself to be impacted.
> 
> Wdyt?

@iarlyy could we preserve the current inheritance by default while allowing users to opt out? For example, `serviceHeadless.annotations: null` could inherit `service.annotations`, `{}` could explicitly disable annotations, and a non-empty map could replace them.

iarlyy:
> @iarlyy could we preserve the current inheritance by default while allowing users to opt out? For example, `serviceHeadless.annotations: null` could inherit `service.annotations`, `{}` could explicitly disable annotations, and a non-empty map could replace them.

@pront good call, I just pushed it.

pront:
@codex review

PR #538, opened 2026-03-05
fix(vector): use haproxy podPriorityClassName in haproxy deployment

kurochan:
Fixes a bug in the vector chart HAProxy Deployment template where priorityClassName was incorrectly read from `.Values.priorityClassName` instead of `.Values.haproxy.podPriorityClassName`.

values.yaml and README document `haproxy.podPriorityClassName` as the HAProxy pod setting, but the template was using the global value.


pront:
@codex review

pront (approved):
Thanks @kurochan

PR #539, opened 2026-03-05
fix(vector): set haproxy terminationGracePeriodSeconds on pod spec

kurochan:
Fixes HAProxy Deployment rendering in the vector chart so `haproxy.terminationGracePeriodSeconds` is applied to the Pod spec.

Previously, `haproxy.terminationGracePeriodSeconds` only affected hard-stop-after in haproxy.cfg, but was not reflected in the Pod spec.
This change makes chart behavior consistent with the value’s intent and documentation.


pront:
@codex review

pront (commented):
Thanks!

pront (approved):

PR #548, opened 2026-03-18
chore(chart): bump minimum kubeVersion to 1.28 and remove legacy version guards

pront:
## Summary

- Bumps `kubeVersion` in `Chart.yaml` from `>=1.15.0-0` to `>=1.28.0-0`, motivated by [vector#24699](https://github.com/vectordotdev/vector/pull/24699) which updated CI test versions from k8s 1.19–1.23 (EOL) to 1.31–1.35
- Removes all version-conditional template branches that existed to support Kubernetes < 1.24 (all long EOL)
- Now evaluating more versions based on  https://github.com/vectordotdev/vector/blob/master/.github/workflows/k8s_e2e.yml#L132-L138

## Why 1.28?

All template simplifications are valid at 1.28 — the oldest removed API (`autoscaling/v2beta2`, `policy/v1beta1`) was dropped upstream in 1.25–1.26, so any cluster on 1.28 was already exclusively using the new APIs. The old branches were dead code for them.

1.28 is also the most conservative defensible floor:

| Provider | 1.28 support ended |
|---|---|
| **GKE** | Feb 2025 |
| **EKS** | Nov 2025 (extended support) |
| **AKS LTS** | Feb 28, 2026 — just expired |
| **OpenShift 4.15** | Aug 2025 (no forced upgrades on-prem) |
| **VMware Tanzu on vCenter 7.x** | May 2025 |

AKS LTS 1.28 is the last major cloud provider to have dropped it, expiring just weeks ago. Anyone below 1.28 is on a version with no active cloud provider support and is very unlikely to be actively updating a Vector chart.

### Version gate (`kubeVersion`)
Helm will **reject `install`/`upgrade`** on clusters running Kubernetes < 1.28. Given 1.27 and below is EOL everywhere, impact should be negligible.

### Template simplifications
**Completely transparent to all users on 1.28+.** Every API that was behind a version guard has been the only available version since well before 1.28:

| Change | API GA since | Old beta/deprecated API removed in |
|---|---|---|
| `networking.k8s.io/v1` for Ingress | k8s 1.19 | beta removed in 1.22 |
| `policy/v1` for PDB | k8s 1.21 | beta removed in 1.25 |
| `autoscaling/v2` for HPA | k8s 1.23 | beta removed in 1.26 |
| `minReadySeconds` on StatefulSet | k8s 1.22 | n/a |
| `persistentVolumeClaimRetentionPolicy` on StatefulSet | k8s 1.23 | n/a |

No behavior change for any user on a supported cluster.

## Test plan

- [x] `helm template` renders correctly against a 1.28+ cluster
- [ ] Verify `helm install`/`upgrade` is rejected on clusters < 1.28 (Helm kubeVersion enforcement)
- [ ] Confirm ingress, statefulset, PDB, and HPA manifests render with the correct API versions

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -42,13 +42,9 @@ jobs:
         # When changing versions here, check that the version exists at: https://github.com/yannh/kubernetes-json-schema and https://hub.docker.com/r/kindest/node/tags
         # Aim to match https://github.com/vectordotdev/vector/blob/master/.github/workflows/k8s_e2e.yml as close as possible
         k8s:
-          - v1.23.3
-          # fails on `kind` but not `minikube`; we weren't able to track down why
-          # possibly related to https://github.com/kubernetes-sigs/kind/issues/2532
-          # - v1.22.4
-          - v1.21.2
-          - v1.20.7
-          - v1.19.11
+          - v1.32.0
+          - v1.30.0
+          - v1.28.0
     steps:
       - name: Checkout
         uses: actions/checkout@v5
@@ -73,13 +69,9 @@ jobs:
         # When changing versions here, check that the version exists at: https://github.com/yannh/kubernetes-json-schema and https://hub.docker.com/r/kindest/node/tags
         # Aim to match https://github.com/vectordotdev/vector/blob/master/.github/workflows/k8s_e2e.yml as close as possible
         k8s:
-          - v1.23.3
-          # fails on `kind` but not `minikube`; we weren't able to track down why
-          # possibly related to https://github.com/kubernetes-sigs/kind/issues/2532
-          # - v1.22.4
-          - v1.21.2
-          - v1.20.7
-          - v1.19.11
+          - v1.32.0
+          - v1.30.0
+          - v1.28.0
     steps:
       - name: Checkout
         uses: actions/checkout@v5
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yaml
```diff
@@ -42,13 +42,12 @@ jobs:
...
         # Aim to match https://github.com/vectordotdev/vector/blob/master/.github/workflows/k8s_e2e.yml as close as possible
         k8s:
-          - v1.23.3
-          # fails on `kind` but not `minikube`; we weren't able to track down why
-          # possibly related to https://github.com/kubernetes-sigs/kind/issues/2532
-          # - v1.22.4
-          - v1.21.2
-          - v1.20.7
-          - v1.19.11
+          - v1.35.0
```
thomasqueirozb, line 45: Is it possible to add a - latest here? That'd be nice

charts/vector/templates/ingress.yaml
```diff
@@ -55,10 +43,6 @@ spec:
                   {{- else }}
                   number: {{ .port.number }}
                   {{- end }}
-              {{- else }}
-              serviceName: {{ if $.Values.haproxy.enabled }}{{ include "haproxy.fullname" $ }}{{ else }}{{ include "vector.fullname" $ }}{{ end }}
```
thomasqueirozb, line 59: Why was this removed?

.github/workflows/ci.yaml
pront: I understand the motivation but I would prefer not based on the "pin every version" principle and to avoid unexpected breakages. Also, this is the pattern we follow in Vector CI (which is what does the heavy testing).

charts/vector/templates/ingress.yaml
pront: The `serviceName/servicePort` fields belong to the pre-1.19 `extensions/v1beta1` Ingress API, which was removed from Kubernetes in `1.22`. Since we're bumping the minimum supported version to `1.28`, this branch was dead code — no cluster we support would ever have hit it.

thomasqueirozb (approved):

PR #583, opened 2026-07-23
fix(releasing): Make post-release steps idempotent and retry-safe

thomasqueirozb:
## Summary

Splits the single `post-release` job into two separate jobs (`sync-to-master` and `bump-version`) so each can be re-run independently if it fails, without re-triggering the other.

Adds idempotency and correctness fixes throughout:

- `sync-to-master` checks for an existing PR (open or merged) before creating one, handles the CLOSED state explicitly, and skips re-enabling auto-merge when it is already active
- `bump-version` checks for an existing open bump PR before creating one, and uses `--force-with-lease` on push to handle retries cleanly
- Both jobs validate the extracted `CHART_VERSION` against a semver regex before proceeding
- `sed` replacement in the version bump now escapes dots in the version string

### DIFF
--- a/.github/workflows/release-post.yml
+++ b/.github/workflows/release-post.yml
@@ -151,13 +151,36 @@ jobs:
       - name: Create and auto-merge PR
         run: |
           BUMP_BRANCH="bump-chart-version-${{ steps.version.outputs.next }}"
-          gh pr create \
+          if ! gh pr create \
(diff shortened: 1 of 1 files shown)


thomasqueirozb:
@codex review

thomasqueirozb:
@codex review

pront (approved):
Discussed offline.

thomasqueirozb:
@codex review

PR #593, opened 2026-08-27
feat(vector): add service loadBalancerSourceRanges field

noamrcon:
Adds `service.loadBalancerSourceRanges` and `haproxy.service.loadBalancerSourceRanges` so a `type: LoadBalancer` Service can be restricted by client IP, matching the existing `loadBalancerIP` field. Defaults to `[]` (unrestricted) and is only rendered when set.

Tested with `helm template`; `helm lint` passes. Supersedes #485.


noamrcon:
Would appreciate a quick look/review!

